PR #32, opened 2026-05-17
docs(changelog): record byte5ai engineering-standards bootstrap

iret77:
## Summary

- Backfills the CHANGELOG entry that should have shipped with PR #31 (byte5ai engineering-standards bootstrap).
- Per AGENTS.md the standards rollout (`.github/engineering-standards.yml`, `.hooks/pre-push`, `script/setup`, AGENTS/CONTRIBUTING updates, branch protection wiring) qualifies as an ops change and required a same-step CHANGELOG note.
- Doubles as the first end-to-end CI verification after GitHub Actions were reactivated on 2026-05-17 (Actions had been disabled since 2026-05-11).

## Test plan

- [ ] All five required status checks reach a terminal state:
  - `middleware (lint + typecheck + test)`
  - `web-ui (lint + typecheck + vitest)`
  - `schema (migrations on pgvector)`
  - `audit (high+critical block) (middleware)`
  - `audit (high+critical block) (web-ui)`
- [ ] CHANGELOG renders correctly on GitHub.

PR #34, opened 2026-05-17
fix(ci): align schema-smoke with current migrations layout

iret77:
## Summary

- The `schema (migrations on pgvector)` job has been failing since Actions were reactivated this morning. The hardcoded `MIGRATIONS` array in `.github/workflows/ci.yml` pointed at a stale layout: four files had moved from `middleware/src/services/graph/migrations/` into the `harness-knowledge-graph-neon` package, and two had been renumbered (`0007->0012`, `0008->0013`) as five new migrations were inserted ahead of them.
- Replaces the hardcoded array with a loop over five migration *domains* (`graph`, `auth`, `routines`, `profileSnapshots`, `profileStorage`). Files inside a domain apply in lexical (numbered) order; domains apply in dependency order.
- Side effect (intentional): the smoke now covers **every** migration in the repo (20 instead of 9), including the previously-untested `auth`, `profileSnapshots`, `profileStorage` domains and the newer routines/graph entries. New migrations will be picked up automatically — no more workflow edits to keep CI honest.

No application code, no migrations, no `package.json` touched — only `.github/workflows/ci.yml`.

## Test plan

- [ ] `schema (migrations on pgvector)` turns green for the first time since Actions were re-enabled.
- [ ] Idempotency re-apply step still passes (each migration is `IF NOT EXISTS`-safe).
- [ ] `web-ui` and `release-blocker` remain green.
- [ ] `middleware` and `audit (middleware)` are expected to keep failing until the parallel Node-20→22 PR lands; that's tracked separately and not in scope here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,9 @@
 # PR + main-branch quality gate.
 #
 # Runs lint + typecheck + test on both workspaces, plus a schema smoke
-# that applies all 9 migrations against a real pgvector container. Audit
-# enforces a high-severity ceiling so a CVE-bumping dep can't slip in
-# unnoticed.
+# that applies every migration in the repo against a real pgvector
+# container. Audit enforces a high-severity ceiling so a CVE-bumping
+# dep can't slip in unnoticed.
 #
 # Naming-independent — no secrets, no GHCR/npm push. Release-flow lives
 # in release.yml (currently placeholder; activates once OB-30 Block 2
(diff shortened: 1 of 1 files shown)


iret77:
Superseded by #35 (consolidated CI resurrection). Under the strict branch protection with `strict: true` required checks, the three PRs (#32/#33/#34) were mutually blocking — each had at least one red required check depending on one of the others. All changes from this PR are preserved in #35 (cherry-picked under the same author identity), plus a fourth commit that fixes the `prefer-const` lint bug that had been hidden behind EBADENGINE, and an extended CHANGELOG `### Fixed` block.

Issue #37, opened 2026-05-17
Repair: middleware test suite (~243 failures, surfaced when Actions were reactivated)


iret77:
## Context

On 2026-05-17, as part of the byte5ai engineering-standards rollout (PR #31), GitHub Actions were re-enabled on this repo (they had been disabled since 2026-05-11). The first complete CI run after reactivation went through on PR #35 (`fix/ci-resurrection`) and exposed **243 test failures out of 2189 tests** (~89% green, ~11% red) in the `middleware (lint + typecheck + test)` job.

The other four required status checks are clean:
- ✓ `web-ui (lint + typecheck + vitest)`
- ✓ `schema (migrations on pgvector)` (now 20 migrations via the glob loop)
- ✓ `audit (high+critical block) (middleware)` (Node 22 fix takes effect)
- ✓ `audit (high+critical block) (web-ui)`

PR #35 stays **open and unmergeable** until this issue is resolved. Standards decision: keep CI strict, accept the merge backlog rather than weaken required checks.

## Failure clusters

Rough classification (source: https://github.com/byte5ai/omadia/actions/runs/25988273696):

| Cluster | ~Count | Nature | Notes |
|---|---|---|---|
| sharp module load on linux-x64 | 41 | Environment bug | "Could not load the sharp module using the linux-x64 runtime" — all `test/diagram*.test.ts` files. `npm ci --no-audit --no-fund` doesn't pull the optional dependency for linux-x64. Fix candidate: `npm ci --include=optional` in `.github/workflows/ci.yml`, or an explicit `npm install --os=linux --cpu=x64 sharp` step. |
| Builder pipeline | ~100 | Real drift in `middleware/src/plugins/builder/*` | BuildPipeline, BuilderAgent, codegen, staging, ESLint persist-back, SlotTypecheckPipeline, buildSandbox, boilerplateSource, etc. |
| Verifier pipeline | ~15 | Drift in `middleware/packages/harness-verifier/*` | `verifier/deterministicChecker` (amount, date, id, graph, aggregate, error handling), `verifier/pipeline` |
| Profile loader / profile routes | ~10 | Drift | `profileLoader / built-in profiles`, `profileLoader / object-form entries with config`, `profileLoader / invalid inputs`, `/api/v1/profiles router` (GET /, /:id, /export, POST /:id/apply) |
| agent-reference | ~5 | Drift | Toolkit `query_notes_by_person`, Store filter `is_reference_only`, SubAgentAccessor |
[...]

PR #35, opened 2026-05-17
fix(ci): resurrect CI after Actions reactivation

iret77:
## Summary

Consolidates the three pre-existing CI bugs that surfaced when GitHub Actions were re-enabled (see PR #31), plus the missed CHANGELOG entry for the standards rollout. Replaces the three separately opened PRs #32, #33, #34, none of which could merge on its own under the strict branch protection — each had at least one red required check dependent on one of the others.

## What's included

| Commit | What |
|---|---|
| `fix(ci): align schema-smoke with current migrations layout` | From #34. Replaces the hardcoded migrations array in `ci.yml` with a glob over five migration domains. Coverage grew 9 → 20 migrations. |
| `fix(ci): bump setup-node to v22 to match middleware engines` | From #33. `node-version: '20'` → `'22'` in three places. Resolves EBADENGINE in `middleware` and `audit (middleware)`. |
| `fix(lint): silence prefer-const on intentional forward reference` | New. `middleware/src/index.ts:398` — eslint doesn't trace the reassignment ~1300 lines later. Single-line disable with explanatory comment. |
| `docs(changelog): record byte5ai engineering-standards bootstrap` | From #32. Initial entry for the standards bootstrap that was missed in PR #31. |
| `docs(changelog): record CI resurrection fixes` | New. `### Fixed` block for the three CI bugs in this PR. |

## Background

- Actions were disabled on this repo since 2026-05-11. Re-enabled today as part of the engineering-standards rollout.
- The first post-reactivation run showed 3/5 required checks red. None of the causes are caused by the standards work — they're drift since the last successful run (migration refactor, Node major bump in `package.json`, a lint false-positive).
- Three individual fix PRs (#33, #34, plus the planned prefer-const PR) could not merge independently: with `strict: true` required checks, each had to wait on the others before being green.

## Test plan

- [ ] All 5 required status checks green on this PR:
  - `middleware (lint + typecheck + test)`
  - `web-ui (lint + typecheck + vitest)`
  - `schema (migrations on pgvector)`
  - `audit (high+critical block) (middleware)`
  - `audit (high+critical block) (web-ui)`
- [ ] Schema smoke now applies 20 migrations (up from 9); idempotency re-apply stays green

## Cleanup after merge

- Close PRs #32, #33, #34 with a pointer back here (already done)
- Remove the three local worktrees + branches under `.claude/worktrees/` (already done)

PR #38, opened 2026-05-17
fix(ci): install sharp's optional linux-x64 binary

iret77:
## Problem

`npm ci --no-audit --no-fund` skips platform-specific optional dependencies. On the Linux-x64 GitHub Actions runner this means sharp's prebuilt binary is never installed, causing 41 diagram test files to bail at startup:

```
Error: Could not load the "sharp" module using the linux-x64 runtime
- Ensure optional dependencies can be installed:
    npm install --include=optional sharp
```

This is cluster 1 of the CI failure investigation tracked in #37 (243 total failures on PR #35, 41 of which are attributable to the missing sharp binary).

## Fix

Add `--include=optional` to all three `npm ci` calls in `.github/workflows/ci.yml` (middleware, web-ui, and audit jobs) so the prebuilt linux-x64 binary lands on disk during CI.

## Expected outcome

This PR should drive the 41 diagram-test failures to zero. The remaining ~200 failures in `middleware (lint + typecheck + test)` are tracked separately in #37 and will be addressed in follow-up cluster fixes.

Only `.github/workflows/ci.yml` is touched — 3 lines changed.

Closes part of #37.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
           cache-dependency-path: middleware/package-lock.json
 
       - name: npm ci (middleware + workspaces)
-        run: npm ci --no-audit --no-fund
+        run: npm ci --include=optional --no-audit --no-fund
 
       - name: Build workspace packages (compiled dist/ for cross-package imports)
         run: npm run build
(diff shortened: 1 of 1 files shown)


Weegy:
Superseded by #44. Main already has `--include=optional` on all 4 `npm ci` calls (vs. 3 here) **plus** an explicit `npm install --no-save --os=linux --cpu=x64 sharp` step that PR #38 doesn't have — needed because `package-lock.json` (generated on macOS) has no `node_modules/@img/sharp-linux-x64` entry so `--include=optional` alone is insufficient. No work lost.

PR #39, opened 2026-05-17
fix(builder): resolve assets and test refs after public-release rename

iret77:
## Summary

Fixes the builder-pipeline cluster from #37 (~100 failures) by repairing the **single release-time root cause** that was misclassified as multi-cluster drift in the original triage.

### Finding

The cluster framing in #37 assumed test-vs-code drift since the 2026-05-11 baseline (`65a70d5`). Empirically wrong: only one commit on `middleware/src/plugins/builder` or `middleware/test/builder` has landed since baseline (`f48cae4`, pure comment translation, net 1:1 line swap). PR #31 touched zero middleware files. The failures predate the public release; CI being disabled until 2026-05-17 hid them.

Reading the actual error output (run 25988273696) instead of the test names reveals two release-time bugs introduced by the byte5→omadia rebrand:

**A — Asset devFallback paths point at a tree excluded from the public release**
`middleware/src/platform/assets.ts` resolved the `boilerplate` and `entityRegistry` bundles via `<repo>/docs/harness-platform/...`. That tree is internal-only. The actual shipped assets are at `middleware/assets/boilerplate/` and `middleware/assets/entity-registry.v1.yaml`. CI sets neither `BUILDER_BOILERPLATE_DIR` nor `BUILDER_ENTITY_REGISTRY_PATH`, so the broken fallback was hit unconditionally and cascaded into every test touching `loadBoilerplate`, `discoverTemplates`, `loadBuildTemplateConfig`, `ensureBuildTemplate`, `BuildPipeline`, codegen, preview routes, the BuilderAgent system-prompt seed, and `SlotTypecheckPipeline`.

**B — One test still asserts the pre-rename package name**
`middleware/test/builder/buildTemplate.test.ts` expected a workspace symlink at `node_modules/@byte5/harness-plugin-api`. The workspace package is `@omadia/plugin-api`; `workspaceDeps` is keyed accordingly. All other call sites already handle both prefixes; this single test missed the rename.

### Changes

- `middleware/src/platform/assets.ts` — both `devFallback` paths and `prodHint` strings (plus their mirrors inside `_rebuildAssetsForTests`) switched from `docs/harness-platform/…` to `middleware/assets/…`. `REPO_ROOT` becomes unused and is removed.
- `middleware/test/builder/buildTemplate.test.ts` — symlink-path constants and the now-dead `assert.match(pkgJson.name, /@byte5\/harness-plugin-api/)` aligned to `@omadia/plugin-api`.

### Expected impact

Drives the bulk of the ~100 builder-pipeline failures from #37 to zero — every test whose stack ends in `loadBoilerplate` (i.e. the entire `boilerplateSource`, `discoverTemplates`, `ensureBuildTemplate`, `loadBuildTemplateConfig`, `BuildPipeline`, `BuilderAgent` system-prompt seed, builder preview routes ENOENT bucket) plus the three `buildTemplate.test.ts` rename failures.

Any residual builder failure that survives the fix is a real next-layer issue that only becomes visible once boilerplate loading actually works — to be triaged in a follow-up PR scoped narrowly.

### Out of scope

- `test/diagram*` — Session 1 / sharp linux-x64 (PR #38).
- `agent-reference` array-order assertions — separate cluster.
- `profileLoader` / profile-routes YAML `@`-char failures — separate cluster.
- `test/agents/resolveAgentForTool.test.ts` — imports `vitest` in a middleware test, separate cluster.
- Plugin registry, manifest, verifier — separate clusters.

### Base branch

Stacked on `fix/ci-resurrection` (PR #35), not `main`. `main` still carries Node-20 + stale migrations + the `prefer-const` lint bug; a CI run on a `main`-based branch would die in EBADENGINE before it could measure this fix. PR #38 (sharp) is stacked the same way. When PR #35 merges, GitHub will rebase this onto `main` automatically.

## Test plan

- [ ] CI green on the new run (or: builder-cluster failures drop from ~100 to a small residual).
- [ ] Verify zero ENOENT failures rooted in `BoilerplateSource: template '<id>' not found`.
- [ ] Verify `buildTemplate.test.ts` symlink + pkgJson assertions pass.

Refs #37, #35.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: middleware/package-lock.json
 
@@ -75,7 +75,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: web-ui/package-lock.json
 
@@ -181,7 +181,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: ${{ matrix.workspace }}/package-lock.json
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,9 @@
 # PR + main-branch quality gate.
 #
 # Runs lint + typecheck + test on both workspaces, plus a schema smoke
-# that applies all 9 migrations against a real pgvector container. Audit
-# enforces a high-severity ceiling so a CVE-bumping dep can't slip in
-# unnoticed.
+# that applies every migration in the repo against a real pgvector
+# container. Audit enforces a high-severity ceiling so a CVE-bumping
+# dep can't slip in unnoticed.
 #
 # Naming-independent — no secrets, no GHCR/npm push. Release-flow lives
 # in release.yml (currently placeholder; activates once OB-30 Block 2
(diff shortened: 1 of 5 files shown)


Weegy:
Superseded by #44. Both unique fixes are already in main:

- **assets.ts devFallback paths** `docs/harness-platform/{boilerplate,entity-registry.v1.yaml}` → `middleware/assets/{boilerplate,entity-registry.v1.yaml}` ([commit f962551](https://github.com/byte5ai/omadia/commit/f962551))
- **buildTemplate.test.ts** stale `@byte5/harness-plugin-api` symlink path → `@omadia/plugin-api` ([commit 0bcb1fd](https://github.com/byte5ai/omadia/commit/0bcb1fd))

The remaining commits on this branch are the common CI fixes already in main (Node 22, schema-smoke migration glob via PR #35, etc.) or obsoleted CHANGELOG entries. No work lost.

PR #43, opened 2026-05-17
fix(verifier): align deterministicChecker and pipeline tests with current API

iret77:
## Summary

Fixes ~17 failures in the verifier-pipeline cluster (Issue #37).

- **Cluster:** Verifier-Pipeline
- **Failures addressed:** ~17 (16 in deterministicChecker, 1 in pipeline)
- **Triage distribution:** All CODE-DRIFT-WEITER — three intentional API changes in the production code that the tests lag

### Root cause: 3 interface changes since tests were written

**1. DeterministicChecker Odoo support removed**

Amount / aggregate / date / id checking against Odoo was extracted from the core checker.
`DeterministicChecker` now handles only `expectedSource === 'graph'`; everything else
returns `unverified` with a message directing operators to install a domain SourceChecker
plugin. The `odoo:` constructor option and `OdooReader` type are gone. 16 test assertions
expected `'verified'` or `'contradicted'` for Odoo-source claims — all now return
`'unverified'`.

Fix: remove `OdooReader` import + `OdooCall` interface + `stubOdoo` helper; update all
Odoo-source assertions to `'unverified'`.

**2. Claim field `odooRecord` renamed to `sourceRecord`**

`HardClaim.odooRecord` was renamed to `sourceRecord` (generic `RecordRef`). Both test
fixtures used the old name, causing `claim.sourceRecord === undefined` in the graph
path, returning `'unverified'` instead of `'verified'`/`'contradicted'`.

**3. `GraphReader` row shape changed: `{ id, displayName }` → `{ id, props }`**

`DeterministicChecker.checkGraph` now reads `row.props?.['displayName']`. The `stubGraph`
helper returned `{ id, displayName }` directly, so the display-name lookup always fell
back to `row.id`, causing the graph verify test to fail.

**4. `traceMissingCallVerdict` is a no-op in core (pipeline test)**

The context-replay test expected `verdict.status === 'blocked'` when
`domainToolsCalled: ['memory']`. The function is intentionally a no-op (comment in
source: "kept as hook so the pipeline shape stays stable when a SourceChecker registry
lands"). Domain-specific replay detection belongs in a SourceChecker plugin. The claim
now reaches `deterministic.checkAll`, which returns `[]`, giving verdict `'approved'`.

## Test plan

- [ ] CI: `verifier/deterministicChecker - amount` all green
- [ ] CI: `verifier/deterministicChecker - aggregate` all green
- [ ] CI: `verifier/deterministicChecker - date` all green
- [ ] CI: `verifier/deterministicChecker - id` all green
- [ ] CI: `verifier/deterministicChecker - graph` all green
- [ ] CI: `verifier/deterministicChecker - error handling` green
- [ ] CI: `verifier/pipeline` all green

Drives ~17 of the ~100 remaining failures to zero. Other clusters tracked in #37.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: middleware/package-lock.json
 
@@ -75,7 +75,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: web-ui/package-lock.json
 
@@ -181,7 +181,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: ${{ matrix.workspace }}/package-lock.json
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,12 @@
 name: CI
 
 on:
+  # Pull-request trigger accepts any base branch so that cluster-fix
+  # PRs stacked on the test-suite resurrection branch (#35 / #37)
+  # still run CI. Once #35 merges and the cluster work is done, this
+  # can return to `branches: [main]`.
   pull_request:
-    branches: [main]
+    branches: ['**']
   push:
     branches: [main]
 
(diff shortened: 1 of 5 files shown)


Weegy:
Closing as **not needed** — verifier tests are already passing in main:

```
$ node --import tsx --test test/verifierDeterministicChecker.test.ts test/verifierPipeline.test.ts
# pass 27
# fail 0
```

Applying this PR's rewrite of `verifierDeterministicChecker.test.ts` (-139/+75) on top of main produces **4 new failures** because the verifier implementation has moved further since the PR was written:
- `unverified` vs `verified`
- `unverified` vs `contradicted`
- error message `'no odoo reader configured'` vs assertion `/no deterministic checker registered for source/`
- verdict `'blocked'` vs `'approved'`

Tests apparently realigned through a different commit. No action needed.

PR #46, opened 2026-05-17
fix(builder): resolve assets and test refs after public-release rename

iret77:
Successor to #39, which was auto-closed when its base branch `fix/ci-resurrection` was merged into `main` via PR #35. Same single commit (now rebased onto current `main`), no functional change. Two empty CI-trigger commits that were only needed before the wider trigger landed (`da31abd`) have been dropped.

## Summary

Fixes the builder-pipeline cluster from #37 (~100 of 243 baseline failures) by repairing the **single release-time root cause** that was misclassified as multi-cluster drift in the original triage.

### Finding

The cluster framing in #37 assumed test↔code drift since the 2026-05-11 baseline (`65a70d5`). Empirically wrong: only one commit on `middleware/src/plugins/builder` or `middleware/test/builder` has landed since baseline (`f48cae4`, pure comment translation, net 1:1 line swap). PR #31 touched zero middleware files. The failures predate the public release; CI being disabled until 2026-05-17 hid them.

Reading the actual error output (run 25988273696) instead of the test names reveals two release-time bugs introduced by the byte5→omadia rebrand + private→public relocation that happened *before* `65a70d5`:

**A — Asset devFallback paths point at a tree excluded from the public release**
`middleware/src/platform/assets.ts` resolved the `boilerplate` and `entityRegistry` bundles via `<repo>/docs/harness-platform/...`. That tree is internal-only. The actual shipped assets are at `middleware/assets/boilerplate/` and `middleware/assets/entity-registry.v1.yaml`. CI sets neither `BUILDER_BOILERPLATE_DIR` nor `BUILDER_ENTITY_REGISTRY_PATH`, so the broken fallback was hit unconditionally and cascaded into every test touching `loadBoilerplate`, `discoverTemplates`, `loadBuildTemplateConfig`, `ensureBuildTemplate`, `BuildPipeline`, codegen, preview routes, the BuilderAgent system-prompt seed, and `SlotTypecheckPipeline`.

**B — One test still asserts the pre-rename package name**
`middleware/test/builder/buildTemplate.test.ts` expected a workspace symlink at `node_modules/@byte5/harness-plugin-api`. The workspace package is `@omadia/plugin-api`; `workspaceDeps` is keyed accordingly. All other call sites already handle both prefixes; this single test missed the rename.

### Changes

- `middleware/src/platform/assets.ts` — both `devFallback` paths and `prodHint` strings (plus their mirrors inside `_rebuildAssetsForTests`) switched from `docs/harness-platform/…` to `middleware/assets/…`. `REPO_ROOT` becomes unused and is removed.
- `middleware/test/builder/buildTemplate.test.ts` — symlink-path constants and the now-dead `assert.match(pkgJson.name, /@byte5\/harness-plugin-api/)` aligned to `@omadia/plugin-api`.

### Measured impact (verified on the PR #39 measurement run [25989938768](https://github.com/byte5ai/omadia/actions/runs/25989938768))

| metric | value |
|---|---|
| Baseline (run 25988273696) | 243 ✖ markers |
| After this fix | 110 ✖ markers |
| Δ | **−133** |

Drives essentially every `BoilerplateSource: template '<id>' not found` ENOENT cascade to zero. Residual builder failures (small — `patchSpecTool`, `workspaceImportResolver`, `B.8-2 manifest-linter integration`, a few `_pendingUserChoice` emissions) only become visible now that boilerplate loading actually works and will be triaged in a follow-up.

### Out of scope

- `test/diagram*` — Session 1 / sharp linux-x64 (PR #38).
- `agent-reference` array-order assertions — separate cluster.
- `profileLoader` / profile-routes YAML `@`-char failures — separate cluster.
- `test/agents/resolveAgentForTool.test.ts` — imports `vitest` in a middleware test, separate cluster.
- Plugin registry, manifest, verifier — separate clusters.

## Test plan

- [ ] CI green on the new run (or: builder-cluster failures drop from ~100 to a small residual).
- [ ] Verify zero ENOENT failures rooted in `BoilerplateSource: template '<id>' not found`.
- [ ] Verify `buildTemplate.test.ts` symlink + pkgJson assertions pass.

Refs #37, #39 (closed-on-base-merge predecessor), #35 (merged).

### DIFF
--- a/middleware/src/platform/assets.ts
+++ b/middleware/src/platform/assets.ts
@@ -25,7 +25,6 @@ import { fileURLToPath } from 'node:url';
 
 const HERE = path.dirname(fileURLToPath(import.meta.url));
 const MIDDLEWARE_ROOT = path.resolve(HERE, '..', '..'); // <repo>/middleware
-const REPO_ROOT = path.resolve(MIDDLEWARE_ROOT, '..'); // <repo>
 
 export interface AssetBundle {
   /** Logical id, e.g. 'boilerplate'. Surfaced in error messages. */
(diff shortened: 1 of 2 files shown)


Weegy:
Thanks for the resubmit — this caught a real bug I missed when I closed #39. PR #44 updated the primary `ASSETS` registry in `assets.ts` but didn't touch the `_rebuildAssetsForTests()` helper's duplicate path config 30 lines below. Test isolation + only-intermittent-failure masked it. Posted a correction comment on #37.

PR #115, opened 2026-05-21
fix(middleware): gate duplicate-capability installs + reconcile DEV_ENDPOINTS_ENABLED on every boot

SecurID:
## Summary

Three independent boot/install issues that together can turn a healthy
kernel into a `process.exit(1)` crashloop after an admin-UI click the
kernel itself accepted.

- **Install-time provides-collision gate.** `InstallService.create()`
  validates `requires` but never `provides`. The admin UI happily
  installs a second plugin claiming `<name>@<major>` already published
  by an active one; the runtime only notices on the next boot, where
  `buildProviderIndex` throws and `main()` exits 1 — no operator-intent
  signal, no graceful recovery. Adds a symmetric \`provides\`-collision
  gate before the existing requires-walk; throws 409
  \`install.capability_already_provided\` with \`{ capability, ownerId }\`.
- **\`bootstrapMemoryFromEnv\` reconcile.** The \`@omadia/memory\` plugin
  stored \`dev_memory_endpoints_enabled\` from \`DEV_ENDPOINTS_ENABLED\`
  exactly once on first-ever boot. Flipping the env later had no effect
  and \`/api/dev/memory\` stayed unmounted. Now reconciled on every boot
  (idempotent, only logs/writes on change). Operator-owned config keys
  (memory_dir, seed_dir, seed_mode) are left untouched.
- **\`/api/dev/*\` publicPaths bypass.** The boot log line
  *\"DEV endpoints enabled at /api/dev — unauthenticated, LOCAL USE ONLY\"*
  was false. \`requireAuth\` mounted at \`/api\` gated every \`/api/dev/*\`
  request with 401. The publicPaths regex list now appends
  \`/^\\/api\\/dev(?:\\/|\$|\\?)/\` when \`DEV_ENDPOINTS_ENABLED=true\` —
  when the flag is off no \`/api/dev/*\` route is mounted at all, so the
  bypass is unreachable. The compose \`127.0.0.1\` port bindings continue
  to prevent non-local exposure.

The plugin source comments in the two knowledge-graph siblings claim
that \"two-active is structurally impossible\" via
\`ctx.services.provide\`. That check runs at activate(), downstream of
\`buildProviderIndex\` — the kernel dies first. The new install-time
gate makes the claim true at the install boundary.

## Reproducer for the crash

1. Bring up the compose stack with \`DATABASE_URL\` set
   (\`docker compose up -d\`). \`@omadia/knowledge-graph-neon\` is
   auto-installed by \`bootstrapKnowledgeGraphFromEnv\`.
2. In the admin UI, install \`@omadia/knowledge-graph-inmemory\`. (Today
   this succeeds — both plugins now sit in \`installed.json\` with
   \`status: active\`.)
3. Restart the middleware container. Kernel logs:
   \`\`\`
   [middleware] fatal startup error: Error: capability 'knowledgeGraph@1' is provided by both
   '@omadia/knowledge-graph-inmemory' and '@omadia/knowledge-graph-neon' — uninstall one
   \`\`\`
4. Compose's \`restart: unless-stopped\` keeps relaunching the same
   broken state.

After this PR, step 2 fails fast with HTTP 409
\`install.capability_already_provided\` and the registry never reaches
the unrecoverable state.

## Test plan

- [x] \`npm test\` — 2421 tests, 0 fail, 7 skipped (was 2414/0/7
      pre-PR; +14 new tests across capabilityResolver / installService /
      bootstrap).
- [x] \`npm run lint\` — clean.
- [x] \`docker compose build middleware && docker compose up -d\` —
      healthy on first boot, idempotent on restart, neon still active,
      \`/api/dev/memory/list\` returns 200 with the seeded tree.
- [x] Manual: install a sibling provider via the admin UI → 409 with
      structured details.

## Out of scope (left for follow-up)

- Boot-time graceful collision handling in
  \`capabilityResolver.buildProviderIndex\` (mark newer entry
  \`errored\` instead of throwing). Helps recover from already-drifted
  installs, but is a behavior change worth its own PR.
- Catalog endpoint marking sibling providers as
  \`install_state: 'incompatible'\` with a reason, so the UI hides the
  install button entirely.
- Updating the misleading \"structurally impossible\" comments in
  \`harness-knowledge-graph-{inmemory,neon}/src/plugin.ts\`.

### DIFF
--- a/middleware/src/index.ts
+++ b/middleware/src/index.ts
@@ -622,6 +622,20 @@ async function main(): Promise<void> {
       // pages like /p/channel-teams/{hub,tab-config} are public-by-
       // design and the reference dashboard is read-only demo state.
       /^\/p\/[^/]+(?:\/|$|\?)/,
+      // Local dev surfaces. The `/api/dev/*` mount itself is conditional
+      // on `DEV_ENDPOINTS_ENABLED=true` further down (see graph +
+      // memory dev routers around the "DEV endpoints enabled" log line)
+      // — when that flag is on the operator has already opted into an
+      // unauthenticated surface and the local Next-UI relies on the
+      // routes being callable without a session cookie. When the flag
+      // is off, no `/api/dev/*` routes are mounted at all, so this
+      // bypass cannot leak anything. When the flag is on AND the stack
+      // is exposed beyond localhost, that is a separate operator
+      // mistake the compose `127.0.0.1` port bindings are designed to
+      // prevent.
+      ...(config.DEV_ENDPOINTS_ENABLED
+        ? [/^\/api\/dev(?:\/|$|\?)/]
+        : []),
     ],
   });
 
--- a/middleware/src/plugins/bootstrap.ts
+++ b/middleware/src/plugins/bootstrap.ts
@@ -267,10 +267,40 @@ async function migrateAnthropicKeyToVault(
  * Idempotent: once the registry entry exists we never overwrite the
  * operator's settings.
  */
-async function bootstrapMemoryFromEnv(deps: BootstrapDeps): Promise<void> {
+export async function bootstrapMemoryFromEnv(deps: BootstrapDeps): Promise<void> {
   const log = deps.log ?? ((m) => console.log(m));
 
-  if (deps.registry.has(MEMORY_TOOL_ID)) return;
+  const desiredDevEndpoints = deps.config.DEV_ENDPOINTS_ENABLED
+    ? 'true'
+    : 'false';
+
+  // Reconcile path: when the plugin is already installed, env-driven flags
+  // still need to track operator intent expressed via container env on
+  // restart. Without this the operator can flip `DEV_ENDPOINTS_ENABLED` in
+  // .env, restart, and the memory plugin's `dev_memory_endpoints_enabled`
+  // stays at its first-boot value forever — because the rest of the
+  // function returns early when the registry entry exists. Other env-
+  // derived config (memory_dir, seed_dir, seed_mode) is operator-owned
+  // and stays untouched.
+  if (deps.registry.has(MEMORY_TOOL_ID)) {
+    const entry = deps.registry.get(MEMORY_TOOL_ID);
+    if (
+      entry &&
+      entry.config?.['dev_memory_endpoints_enabled'] !== desiredDevEndpoints
+    ) {
+      await deps.registry.register({
+        ...entry,
+        config: {
+          ...entry.config,
+          dev_memory_endpoints_enabled: desiredDevEndpoints,
+        },
+      });
+      log(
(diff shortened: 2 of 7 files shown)


Weegy:
## Maintainer merge note

Merged as `988a3de`. Summary of what was done to get this PR landable, and why — for the record.

### 1. Merged `main` into the branch (not rebase)
The branch was `behind` main by 3 commits (#111 proxy rename, #112 UI rename, #113 dynamic allowlist). Resolved with a **merge** rather than a rebase so the original commit `9d73b47` stays intact and attributable — no force-push, no history rewrite on a contributor's fork branch. No conflicts (the 3 main commits touched disjoint files).

### 2. Maintainer fixup commit `35aed19` — JSDoc correctness
The `bootstrapMemoryFromEnv` doc comment still claimed *"once the registry entry exists we never overwrite the operator's settings"*. The new reconcile path **does** overwrite one key (`dev_memory_endpoints_enabled`) on every boot. Pushed an additive `docs(bootstrap):` commit so the comment states the exception explicitly; operator-owned keys (`memory_dir`, `seed_dir`, `seed_mode`) remain untouched as before. Pure doc change — no behaviour impact.

### 3. CI workflow approval
Both workflow runs sat in `action_required` because the author is a `FIRST_TIME_CONTRIBUTOR` (GitHub's standard gate for new contributors). Approved both runs against the final SHA `35aed19` so they ran against the rebased-up-to-date tree.

### Validation
- **CI:** 7/7 required checks green on `35aed19` (middleware + web-ui lint/typecheck/test, both builds, both audits, schema migrations).
- **Local pre-push:** `npm run typecheck` 0 errors, `npm run lint` 0 errors, and the 3 affected test files (`bootstrap` / `capabilityResolver` / `installService`) at **65/65 pass** — including the new collision-gate cases and the edge case "errored plugin nominally providing a capability does NOT trip the gate" (correct: only `status === 'active'` providers reserve a slot).

### Reviewed and accepted
- **Provides-collision gate** — correct and symmetric to the existing `walkCapabilityInstallChain`; malformed capability refs defensively skipped.
- **`DEV_ENDPOINTS_ENABLED` reconcile** — idempotent, only writes/logs on change.
- **`/api/dev/*` publicPaths bypass** — security-sensitive but defensible: `config.ts` defaults the flag to `false`, no `fly.toml` sets it, `compose.yml` binds `127.0.0.1` only, and the regex `/^\/api\/dev(?:\/|$|\?)/` is properly boundary-anchored (no `/api/development` prefix leak). Accepted for the local-dev use case.

### Follow-up (not blocking, optional)
Consider a hard guard so `/api/dev/*` can never be unauthenticated in a deployed environment even if the env flag is set by mistake — e.g. additionally gate on `NODE_ENV !== 'production'`. The PR's own "Out of scope" section already tracks the related boot-time graceful-collision-handling and catalog `install_state: incompatible` items.

Thanks @SecurID — solid first contribution: clear problem statement, concrete reproducer, +14 tests.

PR #277, opened 2026-06-10
feat(omadia-ui): Tier-2 canvas pipeline — skeleton fix, producer tools (rows/charts/choice), typed UI actions, per-user canvas registry

Weegy:
## Scope

Started as the skeleton-composition fix, grew into the full Tier-2 canvas pipeline for Omadia UI: producer tools, typed UI actions, the per-user canvas registry, and the prompt/observability hardening that came out of live testing against the Electron host app.

## Part 1 — Haiku skeleton composition delivers again

Every canvas turn rendered only the static `Working on it…` fallback skeleton. Two stacked root causes, verified against a live deployment:

1. **Lifetime llm budget** — `LlmAccessor.callsUsed` is context-scoped, so `calls_per_invocation: 8` exhausted after 8 calls since boot (manifest cap raised; kernel fix tracked separately).
2. **Markdown-fenced JSON** — Haiku wraps the tree in fences/preambles; `parseResult` now extracts the JSON payload tolerantly.

Composition logs every fallback reason and every model call duration (measured 1.9–3.2 s per call, no repair retry in steady state).

## Part 2 — Producer tools (Tier-3 → canvas data path)

- **`canvas_publish_rows`** — rows land in the skeleton table as a deterministic `surface_patch`. Explicit `data.containerId` is authoritative (resolves tables nested in panes/tabs). Empty `rows: []` is legitimate (empty data set) and still resolves the skeleton loading state.
- **Charts**: rows published against a `chart` container become points (`label` + `value` idiom, fallback to first string/number per row); optional `chartType` (`bar|line|pie`) overrides the skeleton's guess — the publishing turn has seen the data, the skeleton hasn't.
- **`canvas_publish_choice`** — runtime disambiguation: several plausible records → a clickable `choice` element instead of a prose question. The pick returns as the next turn's structured action.
- **Agent-owned row menus**: optional `actions` on `canvas_publish_rows` become `suggestedActions` on the table — context-menu entries that fit the *current* view, decided by the agent, not hardcoded client-side.

## Part 3 — Typed UI actions + targets (protocol 1.0 §5.1 feedback item)

- `ChatTurnInput.action` (`{ type, payload?, target? }`) and `ChatTurnInput.target` are typed SDK fields now; the dispatcher threads `metadata.action` and the turn's `target` for canvas turns — including row-bound **text** turns (beam / context actions), which were previously dropped and forced a disambiguation round-trip (~15–25 s wasted per context click).
- ui-orchestrator hands them to the main turn as `[canvas-action]` / `[canvas-target]` blocks ("act on THAT record, never ask").

## Part 4 — Composer prompt (Tier-2 vocabulary)

- Full core vocabulary (button/choice/input/toggle/form/toolbar/progress/image/tree) with exact shape grammar.
- Rules: disambiguation renders a `choice`; empty data sets keep `rows: []`; chart type is a Tier-2 decision (time series → line, categories → bar, share-of-whole → pie); BE MINIMAL (latency scales with output length).
- Silent-turn rules in `[canvas-context]`: no narration of planning/lookups/tool calls; plain text only (no markdown into canvas values — also stripped defensively in patch composition); one summary sentence after publishing.

## Part 5 — Per-user canvas registry (`@omadia/ui-channel`)

- New wire messages on the authenticated canvas socket (additive v1.1): `canvas_list_get` / `canvas_list_put` ↔ `canvas_list`.
- Registry is keyed by the verified `session.subject` and persisted via the plugin memory store (`canvases/<subject>.json`, new manifest memory permission; volatile in-memory fallback keeps old deployments wire-compatible).
- Entries carry `{ sessionId, title, color }` plus the last server-authoritative `tree` + `revision` (size-capped at 256 KB, sanitised server-side, whitelist-validated client-side before rendering) — canvases materialise on app start across installs.

## Part 6 — Observability

- `[composition] model call attempt N: Xms, Y chars` on every composer call.
- `[patch-composition] skip: <reason>` / `[surface-synthesis] … UNMAPPABLE (…)` — a silently dropped patch reads as "empty canvas" to the user and previously cost a debugging session.

## Part 7 — Canvas lifecycle & data UX (server side of omadia-ui PR #10)

- **`canvas_refresh`** (issue #5, protocol 1.1 additive): deterministic, LLM-free refresh — the client ships `basedOnRevision` + its current tree; Tier 2 re-resolves the data via captured **source recipes** (`tools.invoke`, threaded through the publish sentinel) and answers with ordinary `surface_patch` runs (first publish per container replaces its rows). Refresh recipes are announced as `surface_data_ref_created`; rate-limited to one refresh per canvas session per 3 s; the fallback-refresh prompt asks for the recipe so the NEXT refresh upgrades to the LLM-free path.
- **`turn_abort`** (issue #13): the channel races the orchestrator stream against the abort signal — immediate effect, generator stack unwound without awaiting (a hung tool call cannot delay it), `turn_error 'aborted'`; already-applied surface events stay.
- **Batch rule** for `canvas_publish_rows` + loud non-`end_turn` finalize in the orchestrator.

## Part 8 — Per-user desktop registry (multi-desktop workspaces)

- `desktop_list_get` / `desktop_list_put` ↔ `desktop_list` on the authenticated canvas WS, mirroring the canvas registry: desktops (name, color, `updatedAt` for last-write-wins) with **sessionId-keyed split-tree layouts** so they travel across installs.
- Sanitised server-side (max 24 desktops, layout depth 8, ratio/name/color clamps; an invalid leaf collapses its split); persisted per authenticated subject in the plugin memory store.

## Tests

`uiOrchestratorPlugin` / `uiOrchestratorComposition` / `uiChannelWebSocket` / `orchestratorDispatcher` / `coreApi` extended and green; `tsc` and eslint clean (pre-existing dist-import failures in unrelated packages unchanged). Verified end-to-end against the omadia-test Docker stack with the Electron host app (skeleton → data patch → choice round-trip → registry sync).


Weegy:
Added three canvas-pipeline commits supporting agent-kind plugins that publish their own deterministic canvas trees (first consumer: the X Studio plugin):

- **Sub-agent sentinel synthesis** (`bfea448`): surface synthesis now also scans `sub_tool_use`/`sub_tool_result` (already forwarded by the orchestrator), so a canvas sentinel emitted by a sub-tool inside a domain tool is synthesised — previously only the main-turn tool stream was scanned, so agent-kind plugins could never render constructs.
- **Canvas-output autodiscovery** (`85c843f`): declare → resolve → derive. A plugin declares `canvas_output: true` per manifest capability; the dynamic agent runtime resolves these into a kernel `canvasOutputRegistry` service on (de)activation; the ui-orchestrator derives its sentinel allow-set from it lazily. Replaces the manual `canvas_output_tools` operator field as the primary mechanism (kept as override). Deny-by-default preserved.
- **`tone` schema token** (`4b6de95`): presentational severity hint (neutral/info/success/warning/danger) on commonTraits, mapped client-side to a Lume status colour. Kept in sync with the omadia-ui client schema copy.

All verified live end-to-end against the omadia-test stack (X Studio wizard/drafts/review/generate render; self-review badges show PASS/WARN/FAIL colour).

PR #305, opened 2026-06-15
feat(install): multiline setup fields for string/secret values

SecurID:
## What
Add an optional `multiline: true` flag to `setup.fields` of type `string` or `secret`. When set, the install drawer renders a `<textarea>` instead of a single-line input, so values containing newlines (PEM private keys, certificates) survive intact to the configure endpoint.

## Why
Single-line `<input>` fields silently strip/mangle newlines, so a plugin can't accept a multi-line secret (e.g. an RSA private key) at install time. This is a small, reusable core primitive — any plugin needing a multi-line secret/string can opt in. Older cores ignore the flag and fall back to a single-line input.

## Test plan
- [x] `node --test test/installService.test.ts` in middleware — 16 pass (4 new `extractSetupSchema` multiline cases: passthrough for string/secret, undefined when unset, ignored on non-text types, ignored for non-boolean values)
- [x] `tsc --noEmit` green in both `middleware` and `web-ui`
- [x] manual: install drawer renders the field as a textarea; a 3-line value round-trips to the vault intact; non-multiline fields render unchanged

## Risk / blast radius
- Additive only: new optional manifest field + an extra render branch. No schema/migration changes.
- `PluginSetupField`/`InstallSetupField` gain an optional `multiline?: boolean` — backward compatible (absent → false → today's behavior).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -173,6 +173,14 @@ middleware/packages/harness-integration-confluence/
 middleware/packages/harness-integration-microsoft365/
 middleware/packages/harness-integration-odoo/
 
+# ─── Hub-distributed plugins ──────────────────────────────────────────────
+# GitHub integration + agent. Distributed via the Omadia Hub
+# (hub.omadia.ai), not committed here — the published .zip is the source of
+# truth. Kept on disk as a working-dev copy for local build/test/publish.
+middleware/packages/integration-github/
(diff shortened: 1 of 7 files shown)


SecurID:
Leftover from the old github-integration implementation. But still useful for future plugins if needed.

PR #326, opened 2026-06-17
fix(privacy-guard): render masked columns in canvas_publish_rows datasetId publishes

Weegy:
## Problem

Result tables on Privacy-Shield turns rendered with **empty masked columns** and **no guard badge** (live-verified on Fly: "offene Debitorenrechnungen" → Rechnungsnummer/Kunde blank).

Root cause (proven by repro + Fly logs): the real path is **`canvas_publish_rows`** with a server-resolved `datasetId` — NOT `v4_render_answer` (which #324 fixed). `resolveDatasetForRender` returns the **real** masked values keyed by the dataset's field **path**. But `patchComposition` mapped cells as `resolvedRow[skeletonColumn.fieldKey]`, and the agent's skeleton fieldKeys don't all match the real paths (Odoo `name`/`partner_id`/`invoice_date_due` vs agent's `invoice_number`/`customer_name`/`due_date`) → mismatched columns render empty. Columns that happened to use real Odoo field names (`invoice_date`, `amount_total`) filled fine.

## Fix — generic, at the single data-plane boundary

- **`resolveDatasetForRender`** (`harness-plugin-privacy-guard`) now returns each column's `classification` (`sensitive-masked` | `safe-cleartext`); type added to `PrivacyResolvedDataset` in `plugin-api`.
- **`canvas_publish_rows`** datasetId branch (`omadia-ui-orchestrator`) derives the table columns from the **resolved dataset**: `fieldKey` = authoritative path, `privacy:"guard-protected"` iff `sensitive-masked`, humanized-path label fallback. Emitted as `data.columns`.
- **`patchComposition`** uses those columns when present: maps cells by the real paths (masked values now populate), preserves the agent's human labels **by position**, replaces the table columns, carries the privacy flag. Odoo many2one `[id,"Name"]` tuples render the display name (not raw JSON). Agent-authored (non-datasetId) publishes are unchanged.

This fixes empty masked cells + the missing badge for **all** datasetId-based canvas producers, not just one path.

## Pairs with
omadia-ui renderer (`feat/privacy-guard-result-display`): shield badge in the column header + violet cell highlight on `column.privacy`. Schema `tableColumn.privacy` enum merged in #324.

## Tests
- New `canvasPublishMaskedColumns.test.ts`: masked column shows the real value, carries `guard-protected`, agent label preserved by position, extra dataset column uses humanized label.
- 209 relevant regression tests (uiOrchestrator/canvas/privacyV4) green; full build clean (Node 22).

## Residual risk (flagged for live verification)
Label reconciliation is **by position** (dataset column i ↔ skeleton column i). FieldKeys/values are always correct (keyed by path); only the human *label* mapping is order-sensitive. Correct for the proven Odoo case; one live canvas check on the masked-invoice query recommended (this PR + Fly deploy → local UI re-test).

### DIFF
--- a/middleware/packages/harness-plugin-privacy-guard/src/service.ts
+++ b/middleware/packages/harness-plugin-privacy-guard/src/service.ts
@@ -404,7 +404,11 @@ export function createPrivacyGuardService(deps?: {
       );
       return {
         rowCount: dataset.rows.length,
-        columns: dataset.schema.fields.map((f) => ({ path: f.path, type: f.type })),
+        columns: dataset.schema.fields.map((f) => ({
+          path: f.path,
+          type: f.type,
+          classification: f.classification,
+        })),
         rows: dataset.rows as ReadonlyArray<Record<string, unknown>>,
       };
     },
--- a/middleware/packages/omadia-ui-orchestrator/src/patchComposition.ts
+++ b/middleware/packages/omadia-ui-orchestrator/src/patchComposition.ts
@@ -43,9 +43,15 @@ export interface ComposedPatch {
 interface TableNode {
   type: 'table';
   loading?: string;
-  columns?: Array<{ fieldKey: string; label: string }>;
+  columns?: Array<{ fieldKey: string; label: string; privacy?: 'guard-protected' }>;
   rows: Array<{ rowKey: string; cells: Record<string, unknown> }>;
   [key: string]: unknown;
+}
+
+interface PublishedTableColumn {
+  fieldKey: string;
+  label: string;
+  privacy?: 'guard-protected';
 }
 
 function escapePointerSegment(seg: string): string {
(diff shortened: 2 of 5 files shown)


Weegy:
**Superseded by #743 — closing.**

Your diagnosis was right and it still holds. What did not survive is the code: this PR was written in June against the pre-v4 privacy guard, and all four production files it touches have moved since Privacy Shield v4 landed. Rebasing would have imported the old architecture's assumptions, so #743 rebuilds the fix against v4 and keeps this PR's diagnosis and test cases.

**The defect was verified before anything was changed** — #743's first commit reproduces it on today's `main`:

- `PrivacyResolvedDataset.rows` is documented as *"the full real rows, keyed by column `path`"*
- `handleCanvasPublishRows` passed them through untouched
- `composeStructuredPayloadPatch` read each cell as `row[fieldKey]`, using the **skeleton's** keys

So `invoice_number` against Odoo's `name` missed and the cell was silently blanked, while columns that happened to reuse the real field name filled correctly — exactly the partial blanking reported here.

Two things #743 adds beyond this PR:

1. **The residual risk you flagged is now bounded.** Labels are still matched positionally — there is no field-key↔path mapping anywhere in the system to do better — but the guess is taken **only when the two column lists are the same length**, and the order dependency is pinned by a test so it cannot drift silently.

2. **The producing service is tested.** A mutation run showed that deleting `classification: f.classification` from `resolveDatasetForRender` failed no test at all: the renderer tests feed a schema straight into the composer and never execute the service. That gap is closed against the real service.

Thanks — the hard part here was the diagnosis, and it was already done.

PR #342, opened 2026-06-19
fix(desktop): installer CI build/sign pipeline + review hardening

Weegy:
## What

Follow-up to #341 (native desktop installer): makes the installer **actually build, sign, and ship** from CI, plus a round of review hardening. The desktop app itself already landed in #341 — this is the delivery pipeline + fixes found by a third review pass and the first real cross-platform build.

## Delivery pipeline

- **`desktop-apps.yml` now builds end-to-end on macOS + Linux** (verified green on the runners; Windows is `continue-on-error` for now, see below). Fixes from the first real cross-platform run: synced lockfile, `@electron/rebuild` pinned to 3.6.1 (avoids a git+ssh node-gyp dep), Python 3.11 (node-gyp needs `distutils`, gone in 3.12), deb maintainer email + project homepage, run-artifact upload so green legs are downloadable, and a 30-min job timeout.
- **`build/` → `buildResources/`** — the repo-root `.gitignore` `build/` rule had silently excluded the signing hook (`afterPack.js`) + entitlements, so they were never committed and electron-builder couldn't find them.
- **Auto-build on release**: a GITHUB_TOKEN-created release does not fire the `release` event, so `desktop-apps.yml` is now also a `workflow_call` invoked **in-run from `auto-release.yml`** (same pattern as `publish-images`). Future releases build + attach the installers automatically.

## Review hardening (3rd pass)

- embeddedDb: close the embedded Postgres if the socket server fails to start (no leaked dataDir lock on retry).
- secrets: only cache the vault key after `persist()` succeeds (don't bypass fail-closed).
- supervisor: a boot superseded by a concurrent stop/restart no longer double-tears-down (generation guard); dropped a dead `readSetup()` on the boot path.
- updater: quiesce/stop the DB before snapshotting it, and step aside in `before-quit` during an install so the installer isn't bypassed.

## Status / known gap

- ✅ **macOS (Apple Silicon), Windows and Linux** all build green. Windows was solved by treating `argon2`/`sharp` as N-API (ABI-stable, no rebuild) + a `better-sqlite3` **Electron prebuild** (no node-gyp/VS); a CI smoke-load now proves all three load under Electron's ABI.
- ⚠️ **macOS is arm64-only for v1.** Native modules ship as extraResources (copied verbatim, not arch-split) and GitHub's Intel (`macos-13`) runners queue for hours, so an x64 DMG is impractical and an arm64 runner would put arm64 modules in an x64 DMG (crash on Intel). Intel x64 is a tracked follow-up (cross-compile x64 prebuilds into a second staged tree).
- Signing stays fail-soft: unsigned artifacts ship until the Apple/Windows secrets are added to the repo.

## Notes for reviewers

- The diff is 12 files vs current `main` (#341 already merged the desktop app).
- `auto-release.yml` gains one job mirroring `publish-images`.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -99,3 +99,16 @@ jobs:
     with:
       ref: ${{ needs.release.outputs.new_tag }}
       version: ${{ needs.release.outputs.new_tag }}
+
+  # Build + attach the native desktop installers (macOS/Linux signed if secrets
+  # are set; Windows best-effort). Same in-run pattern as publish-images, because
+  # the GITHUB_TOKEN-created release does not fire a `release` event.
+  desktop-apps:
+    needs: release
+    if: ${{ needs.release.outputs.new_tag != '' }}
+    permissions:
+      contents: write
+    uses: ./.github/workflows/desktop-apps.yml
+    with:
+      tag: ${{ needs.release.outputs.new_tag }}
+    secrets: inherit
--- a/.github/workflows/desktop-apps.yml
+++ b/.github/workflows/desktop-apps.yml
@@ -52,8 +52,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # macOS is built PER ARCH on a matching-arch runner, because the native
+          # modules ship as extraResources (copied verbatim, not arch-split): a
+          # single arm64 runner would otherwise put an arm64 better_sqlite3.node
+          # into the x64 DMG and crash on Intel Macs. macos-latest = Apple Silicon,
+          # macos-13 = Intel.
           - os: macos-latest
-            args: --mac
+            args: --mac --arm64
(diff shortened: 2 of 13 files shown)


Weegy:
Closing as superseded.

Every file this PR touches now exists on `main` with the work landed through other PRs: #558 (macOS signing — the unsigned-app + symlink defects), #574 (Intel/x64 build) and #675 (stop source-building better-sqlite3 in the desktop build). `.github/workflows/desktop-apps.yml` on `main` carries 47 signing/notarization references today; this branch was last touched 2026-06-19, before any of that.

Rebasing a June installer-pipeline branch onto that would be a rewrite, not a merge. Windows signing — the one genuinely open piece of this area — is tracked separately in #559.

Reopen if a specific hunk here turns out not to be covered.

PR #356, opened 2026-06-22
feat(desktop): native one-click installer — bundled PostgreSQL 17 + pgvector, macOS/Linux/Windows

Weegy:
Supersedes the earlier (closed) #355 — same branch, now also carrying the cross-platform matrix and the onboarding-wizard fix.

## What

Ships the **native one-click desktop installer** for omadia (macOS / Linux / Windows) — an Electron shell that bundles and supervises the existing middleware kernel + web-ui plus a **bundled native PostgreSQL 17 + pgvector**, with **no Docker** and no separate Node runtime. Once merged, `auto-release.yml` builds and attaches the installers to every release automatically (the `desktop-apps` reusable-workflow job).

## Why native Postgres (not PGlite)

The first iteration embedded PGlite (Postgres-in-WASM) over `pglite-socket`. It booted, but the WASM engine crashed (`RuntimeError: unreachable`) under the kernel's real query load and is single-connection. This replaces it with a real bundled PostgreSQL 17 driven directly via `initdb`/`postgres` child processes (loopback-TCP, trust auth), staged as `extraResources` so it is executable on disk (never trapped in the asar).

## Platforms — all three CI-green

| OS | pgvector source | Verification |
|----|-----------------|--------------|
| macOS arm64 | Homebrew `postgresql@17` | build + **boot-tested** end-to-end (auth/health 200, `CREATE EXTENSION vector` 0.8.3 loads, port-drift recovery) |
| Linux x64 | pgdg apt `postgresql-17-pgvector` | build + artifact structurally verified (vector.so, relative symlinks) |
| Windows x64 | MSVC `nmake` build from pgvector source (Chocolatey PG17 for headers) | build + artifact structurally verified (vector.dll PE32+ x64) |

## Notable fixes found by booting real artifacts (not just green CI)

- **Port/vault drift**: the kernel read `database_url` only from the vault (frozen at first boot); a changed embedded-DB port would crash-loop boot. Gated `OMADIA_EMBEDDED_DB=1` makes the live env DSN authoritative on desktop (cloud unchanged). Proven: forced the frozen port busy → app drifted to a new port → graph reconnected.
- **Dangling dylib/so symlinks**: `cpSync({dereference:true})` rewrote the engine's relative library symlinks to absolute build-machine paths (resolved locally, dangled on CI / end-user machines → `dyld: Library not loaded`). Fixed by restoring relative in-tree symlinks from the engine's `pg-symlinks.json` manifest.
- **Onboarding wizard broken on all platforms** (reported on Windows: key test stuck on "Testing…", folder picker dead, setup stalled): the `sandbox: true` preload could not `require('./ipcTypes')`, so `window.omadia` was never exposed and every bridge call silently hung. Fixed by bundling the preload self-contained with esbuild, plus renderer guards so a bridge failure surfaces instead of hanging. Verified on the packaged wizard via CDP (`window.omadia` present, `testLlmKey` round-trips and resolves).
- **Install verbosity**: kernel/web-ui/Postgres output (which all funnel through `log`) is now mirrored to the wizard + loading UI via a gated `bootLog` channel — live scrolling log + elapsed-time counter so a long first boot (migrations) visibly stays alive.

## Signing (fail-soft)

Unsigned/ad-hoc without secrets; the build still produces installers. macOS reuses the proven `_HIGH5` Apple secret names (Developer ID + notarization). Windows Authenticode is optional (`WINDOWS_CSC_*`).

## Known limitations / follow-ups

- **Runtime** boot of the Linux AppImage and Windows installer on those actual OSes is not yet done (only macOS can be booted on the dev box). The bundling is verified; the `vector.so`/`vector.dll` load and the wizard are the same PG17-ABI / sandbox-preload mechanisms proven on macOS.
- macOS is **arm64 only** for v1 (Intel runners queue for hours). Intel x64 is a tracked follow-up.
- Signed/notarized artifacts require the Apple/Windows signing secrets on the repo.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -99,3 +99,16 @@ jobs:
     with:
       ref: ${{ needs.release.outputs.new_tag }}
       version: ${{ needs.release.outputs.new_tag }}
+
+  # Build + attach the native desktop installers (macOS/Linux signed if secrets
+  # are set; Windows best-effort). Same in-run pattern as publish-images, because
+  # the GITHUB_TOKEN-created release does not fire a `release` event.
+  desktop-apps:
+    needs: release
+    if: ${{ needs.release.outputs.new_tag != '' }}
+    permissions:
+      contents: write
+    uses: ./.github/workflows/desktop-apps.yml
+    with:
+      tag: ${{ needs.release.outputs.new_tag }}
+    secrets: inherit
--- a/.github/workflows/desktop-apps.yml
+++ b/.github/workflows/desktop-apps.yml
@@ -3,8 +3,8 @@
 # Release that triggered the run.
 #
 # This is the no-Docker, one-click installer in `desktop/`: an Electron shell
-# that bundles + supervises the existing middleware kernel and web-ui plus an
-# embedded Postgres (PGlite). See desktop/README.md and
+# that bundles + supervises the existing middleware kernel and web-ui plus a
+# bundled native PostgreSQL 17 + pgvector. See desktop/README.md and
 # docs/plans/native-installer-plan.md.
 #
 # Runs separately from release.yml (the GHCR image pipeline) so a desktop build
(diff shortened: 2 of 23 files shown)


Weegy:
Closing — this branch diverged after #355 was squash-merged into main (so it re-applies already-merged commits and conflicts). The remaining work (the onboarding-wizard preload fix + install verbosity) is in a clean PR off current main.

PR #366, opened 2026-06-24
fix(ui): update table rendering behavior

### LINKED ISSUE #319: Rendered Table goes out of bounds
Image

Possible approach: Make tables which are too large to be collapsible to the sides.

sneumannb5:
## What
  GFM tables stay contained: keyboard-focusable scroll wrapper + sticky <thead>. MarkdownTable.tsx plugs into the <table> slot in Markdown.tsx. Closes #319 
                                                                                                                                                
 ## Why                                                                                                                                           
  Wide/tall tables broke chat-bubble layout. Wrapper scoped to .md-table-wrap — BuilderMarkdown and other .md-view surfaces stay at their       
  previous default.                                                                                                                             
                                                                                                                                              
  ## Test plan                                                                                                                                     
  - [x] npm run lint && npx tsc --noEmit (web-ui)                                                                                                 
  - [x] vitest run — Markdown.test.tsx green (was red on a4b9d73)
  - [x] 20-col table → horizontal scroll inside bubble           
  - [x] 200-row table → sticky header keeps its divider while scrolling                                                                             
  - [x] Tab into wrapper → arrow-key scroll (WCAG 2.1.1)                                                                                            
  - [x] Streaming mid-table → no flicker                                                                                                            
  - [x] de + en render markdownTable.scrollRegionLabel                                                                                              
  - [x] BuilderMarkdown small table → content-width, no stretch                                                                                     
                                                                                                                                                
 ## Risk                                                                                                                                          
  Pure web-ui/. One new i18n key. No schema / API / env. Sticky border switched to border-collapse: separate + inset-shadow divider           
  (sticky-collapse bug).                                                                                                                        
                                                                                                                                              
 ## Changes since a4b9d73                                                                                                                         
  - Dropped fullscreen modal (no portal, no framer-motion, no lucide-react, no maximize button) → resolves modal-a11y / focus-trap / inert /  
  motion-token / print-button / button-overlap by elimination.                                                                                  
  - Markdown.test.tsx → renderWithIntl (CI blocker).                                                                                          
  - Dead i18n keys removed (ariaCloseBackdrop, openFullView, modalTitle, ariaClose).                                                            
  - min-width: max-content scoped to wrapper (BuilderMarkdown regression).                                                                      
  - Sticky divider via inset shadow (sticky+collapsed-border bug).                                                                              
  - Wrapper tabIndex={0} + role="region" + aria-label.


web-ui/app/_components/MarkdownTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const onOpen = useCallback(() => {
+    setOpen(true);
+  }, []);
+  const onClose = useCallback(() => {
+    setOpen(false);
+  }, []);
+
+  return (
+    <div className="md-table-block relative">
+      <button
```
ConnysCode: **Floating button — positioning & contrast.**

- The button is `absolute right-1 top-1` on `.md-table-block` (no margin), but the visible table is inside `.md-table-wrap`, which has `my-3` (12px top margin) and is a *sibling* of the button — so the button floats ~8px above the table's top edge in the margin gap rather than on the corner. Move it inside `.md-table-wrap` (already `position: relative`), or shift the spacing onto `.md-table-block`.
- It's `z-[2]` over the sticky header (`z-index: 1`), so it permanently overlaps the top-right `<th>`, and `bg-[color:var(--bg-soft)]/90` over a `--bg-soft` header gives the 14px icon very low contrast. Consider show-on-hover/focus (keep it keyboard-reachable) or a solid chip + right padding on the header row.

web-ui/app/_components/MarkdownTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+      window.removeEventListener('keydown', onKey);
+    };
+  }, [open, onClose]);
+
+  if (!mounted) return null;
+
+  return createPortal(
+    <AnimatePresence>
+      {open ? (
+        <motion.div
```
ConnysCode: **Modal a11y — no focus trap / restore / `inert`; prefer `aria-labelledby`.**

`autoFocus` on the close button is a start, but: (1) Tab isn't trapped, so focus moves into the page *behind* this opaque full-screen overlay; (2) focus isn't restored to the maximize button on close (it falls to `<body>`); (3) the rest of the app isn't `inert`/`aria-hidden`, so screen-reader users can navigate behind the dialog. Setting `inert` on the app root while the modal is open solves (1) and (3) together.

Also, `aria-label={t('modalTitle')}` duplicates the visible `<h3>` on lines 122–124 — give the `<h3>` an `id` and use `aria-labelledby` instead (cf. `ConfirmDialog.tsx`).

web-ui/app/_components/MarkdownTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+}
+
+// ---------------------------------------------------------------------------
+// Motion variants — pure opacity fade. Avoids `transform` on the panel so
+// nothing inside the fullscreen pane inherits a stacking-context that would
+// re-anchor a hypothetical nested `position: fixed` child.
+// ---------------------------------------------------------------------------
+
+const EASE_OUT: [number, number, number, number] = [0.22, 0.61, 0.36, 1];
```
ConnysCode: **Duplicated motion tokens + `prefers-reduced-motion` ignored.**

`EASE_OUT` / `TRANSITION_FAST` / `FADE_VARIANTS` are copied verbatim from `AgentDetailsModal.tsx`. The same curve is already the canonical token in `app/_lib/design-tokens.ts` (`motion.standard: 'cubic-bezier(0.22, 0.61, 0.36, 1)'`) and the CSS var `--easing-standard` — import it instead of adding another copy.

This framer-motion fade also ignores `prefers-reduced-motion`: the global reduced-motion CSS only zeroes CSS transitions, not framer-motion's JS animation. The repo already uses `useReducedMotion()` in `ui/Button.tsx` — reuse it, or wrap the app in `<MotionConfig reducedMotion="user">`.

web-ui/app/globals.css
```diff
@@ -480,18 +480,52 @@ body {
...
   @apply px-3 py-2 text-left align-top tabular-nums;
   border: 1px solid var(--border);
+  overflow-wrap: anywhere;
+  word-break: break-word;
 }
 .md-view th {
   @apply font-semibold;
   background: var(--bg-soft);
 }
+.md-view thead th {
```
ConnysCode: **Sticky header border drops out under `border-collapse`.**

`.md-view table` is `border-collapse` (line 500), so cell borders belong to the table, not the `<th>`. With `position: sticky` on `thead th`, the collapsed top/bottom border doesn't travel with the sticky cell (known Chromium/WebKit behaviour) — the header looks borderless while scrolling and rows sit flush against it. Switch to `border-collapse: separate; border-spacing: 0` and give the header its line via `box-shadow: inset 0 -1px 0 var(--border)` (and `inset 0 1px 0` for the top edge).

web-ui/app/globals.css
```diff
@@ -480,18 +480,52 @@ body {
   @apply font-semibold text-[color:var(--fg-strong)];
 }
 
+/* Tables live inside a scroll-wrapper (MarkdownTable) so very wide tables
+   scroll horizontally inside the chat bubble instead of bleeding past it.
+   The wrapper also gives a max-height + sticky thead for long tables. The
+   `min-width: max-content` keeps columns from being squeezed to unreadable
+   widths — the wrapper takes the scroll instead. */
+.md-table-wrap {
```
ConnysCode, line 488: **Scroll region isn't keyboard-scrollable.**

`.md-table-wrap` is a plain `<div>` with `overflow: auto` and no `tabindex`, so keyboard-only users can reach the maximize button but can't scroll the clipped region (60vh vertical / horizontal) with the arrow keys — content past the fold is then only reachable via the modal. Add `tabindex={0}` + `role="region"` + an `aria-label` on the wrapper so it's focusable and arrow-scrollable (and gets a focus ring).

web-ui/app/globals.css
```diff
@@ -480,18 +480,52 @@ body {
...
 .md-view th {
   @apply font-semibold;
   background: var(--bg-soft);
 }
+.md-view thead th {
+  position: sticky;
+  top: 0;
+  z-index: 1;
+}
+@media print {
```
ConnysCode, line 545: **Print — the floating button isn't hidden.**

This query de-stickies the header and restores overflow (good), but it doesn't hide the `z-[2]` maximize button, so it prints as a grey square over the table's top-right corner. Add `.md-table-block button { display: none }` inside this `@media print` block.

web-ui/messages/en.json
```diff
@@ -273,6 +273,12 @@
     "outputLabel": "Output",
     "subIterations": "{count} inner iterations"
   },
+  "markdownTable": {
+    "openFullView": "Open table in full view",
+    "modalTitle": "Table — full view",
+    "ariaClose": "Close",
+    "ariaCloseBackdrop": "Close"
```
ConnysCode: **Dead i18n key.**

`ariaCloseBackdrop` isn't referenced anywhere in `MarkdownTable.tsx` — this modal has no backdrop button (the panel fills the viewport); it looks copied from the `agentDetailsModal` namespace. Remove it here and in `de.json` (line 280).

Note: with this key the PR adds *four* `markdownTable` keys, not the "two" stated in the description, and the i18n-parity test won't flag the unused one (it only checks en↔de parity).

ConnysCode (changes requested):
**Review: Request changes**

Solid core idea — the scroll-wrapper genuinely fixes the out-of-bounds overflow from #319, and the body-portal for the full-view modal is the *correct* call (assistant bubbles keep a residual `transform` from the `lume-condense` animation via `animation-fill-mode: both`, which would otherwise capture a non-portaled `position: fixed`). A few things to address before merge, one blocking.

### 🔴 Blocker — breaks an existing test (CI red)

Rendering a GFM table now mounts `MarkdownTable` → `useTranslations('markdownTable')`. `app/_components/__tests__/Markdown.test.tsx` renders a table with a plain `render()` (no `NextIntlClientProvider`), so it now throws. Verified locally on head `a4b9d73`:

```
FAIL  Markdown.test.tsx > wraps a highlightTerm occurrence in a violet span
Error: Failed to call `useTranslations` because the context from
       `NextIntlClientProvider` was not found.
 ❯ MarkdownTable app/_components/MarkdownTable.tsx:31:13
Tests  1 failed | 3 passed (4)
```

Fix: migrate the test to the repo's `renderWithIntl` (`app/_lib/test-utils.tsx`), and run the full `vitest run` (the lint/tsc checkbox is unticked).

### 🟠 Scope & architecture

- **Scope creep vs #319.** The issue asks for tables "collapsible to the sides" — the scroll wrapper delivers that. The fullscreen modal + maximize button + sticky thead are extra surface nobody requested; consider shipping the wrapper alone and splitting the modal off.
- **App-wide blast radius — not just chat.** `<Markdown>` renders in ≥6 non-test places (`app/page.tsx`, `app/memory/page.tsx`, `app/routines/[id]/runs/[runId]/page.tsx`, `CreateIssueButton.tsx`, `store/InstallButton.tsx`, `app/store/[id]/page.tsx`). The `table` override + the global `.md-view table` rules change **every** markdown table app-wide; a floating "open fullscreen" button now shows on e.g. static store-description tables. Scope it to chat, or own the app-wide change in the description.
- **Fourth hand-rolled modal.** Duplicates the dialog pattern already in `AgentDetailsModal.tsx`, `OnboardingModal.tsx`, `ConfirmDialog.tsx` (portal + `role="dialog"`/`aria-modal` + ESC + header + X). `app/_components/ui/` is the primitives home (it has `Button.tsx`) but no `Modal`. A shared `ui/Modal` would stop the divergence (focus handling already differs across these four).

### 🧪 Tests

No test for the new ~160-LoC component despite a strong vitest convention; a smoke test (open / ESC / portal target) is cheap. (The existing table test must be fixed regardless — see Blocker.)

### ⚡ Perf framing

The hoisted components object is fine, but with the modal open mid-stream the table `children` live in two DOM subtrees and reconcile twice per token; `useTranslations` runs twice per table. Minor, but it cuts against the "streaming-cheap" framing.

| Claim | Reality |
|---|---|
| "Adds two i18n keys" | Four keys per locale; `ariaCloseBackdrop` is unused (see inline) |
| "tables in chat bubbles" | Applies app-wide (6 `<Markdown>` call-sites) |
| Test-plan checkboxes | All unticked; the table test is in fact red |
| "no schema / API / env-var touch" | ✅ accurate |
| "framer-motion + lucide-react already in deps" | ✅ accurate |

### Minimum to merge

1. Fix `Markdown.test.tsx` (→ `renderWithIntl`) and get CI green.
2. Remove the dead `ariaCloseBackdrop` key; correct the description (four keys, app-wide).
3. Decide chat-only vs app-wide and scope accordingly.

Strongly recommended: address the modal a11y (focus trap / restore, `inert`), the sticky-header border bug, and pull the easing from `design-tokens.ts`. Details inline.

- `globals.css:508-509` — `overflow-wrap: anywhere` + `word-break: break-word` are redundant; the latter is the legacy form and, with `tabular-nums`, can break long numbers mid-digit. `overflow-wrap: break-word` alone is gentler.
- `globals.css:501-502` — `min-width: max-content` + `width: 100%` stretches even small 2-column tables to full width everywhere markdown renders.
- `MarkdownTable.tsx:32` — `open` state survives streaming only for append-only growth; a block-structure change *above* the table mid-stream can remount `MarkdownTable` and silently close the modal. Worth scoping the "streaming-safe" claim.
- `MarkdownTable.tsx:110` — the `motion.div` has no `key`, and the exit fade is dropped if `MarkdownTable` unmounts while open (inherent to a per-instance portal).
- `MarkdownTable.tsx:97` — the ESC handler doesn't `preventDefault`/`stopPropagation`, so it propagates to any ancestor keydown handler.
- JSDoc blocks are heavier than the surrounding code's norm (harmless).
</details>

*(Reviewed against head `a4b9d73`.)*

web-ui/app/globals.css
```diff
@@ -480,18 +480,52 @@ body {
...
+}
+.md-table-wrap--full {
+  max-height: none;
+  border: 0;
+  border-radius: 0;
+}
 .md-view table {
-  @apply my-3 w-full border-collapse text-xs;
+  @apply border-collapse text-xs;
+  min-width: max-content;
```
Weegy, line 519: Regression risk: this changes the **global** `.md-view table` rule, but the scroll wrapper that compensates for `min-width: max-content` is only applied by `Markdown.tsx` via the new `table` component override.

`app/store/builder/[id]/_components/BuilderMarkdown.tsx` is a separate `react-markdown` instance that also renders into `<div className="md-view">` with `remark-gfm`, but it only overrides `ol` (no `table` override, no wrapper). So its tables now expand to `max-content` with no scroll container and can overflow the builder workspace, where previously they were contained at `width: 100%`.

Suggestion: scope the new behavior to the wrapper instead of the global selector, e.g. `.md-table-wrap table { min-width: max-content }`, and keep `.md-view table { width: 100% }` as the default. (`CreateIssueButton` is fine since it delegates to `<Markdown>`.)

web-ui/app/globals.css
Weegy: `position: sticky` on `thead th` combined with `border-collapse: collapse` (set on `.md-view table`) is a known browser issue: with collapsed borders the border belongs to the table, not the cell, so it does not travel with the sticky header and visually disappears while scrolling inside the 60vh box.

Worth verifying in-browser; if it reproduces, switch to `border-collapse: separate; border-spacing: 0` and render the header divider via `box-shadow` on `thead th`.

web-ui/app/_components/MarkdownTable.tsx
Weegy: This button is absolutely positioned at `right-1 top-1` over the table's top-right corner, so it overlaps the last column's header cell on most tables (covering text and intercepting clicks on that cell). It is also rendered for every table, including small ones that never overflow.

Consider only showing it when the table actually overflows (compare `scrollWidth`/`clientWidth` via a ref or ResizeObserver) and/or insetting it into a gutter so it never covers cell content.

web-ui/app/_components/MarkdownTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+        title={t('openFullView')}
+        className={cn(
+          'absolute right-1 top-1 z-[2] rounded p-1',
+          'bg-[color:var(--bg-soft)]/90 text-[color:var(--fg-muted)]',
+          'transition-colors hover:bg-[color:var(--bg-elevated)] hover:text-[color:var(--fg-strong)]',
+        )}
+      >
+        <Maximize2 size={14} aria-hidden />
+      </button>
+      <div className="md-table-wrap">
```
Weegy: Accessibility: this wrapper is `overflow: auto` but not keyboard-focusable, so keyboard-only users cannot scroll wide/tall tables (WCAG 2.1.1 Keyboard, 1.4.10 Reflow). Add `tabIndex={0}`, `role="region"` and an `aria-label`; the repo already styles `[tabindex]:focus-visible`. The same applies to the `md-table-wrap--full` wrapper in the modal.

web-ui/app/_components/MarkdownTable.tsx
Weegy: The dialog sets `aria-modal="true"`, auto-focuses the close button and handles ESC (good), but there is no focus trap, no body scroll lock, and focus is not restored to the trigger on close. Tab can move into the still-present background document (breaking the `aria-modal` contract), and the background scrolls behind the overlay.

Consider a small focus trap, `overflow: hidden` on `body` while open, and restoring `document.activeElement` on close (WCAG 2.4.3).

web-ui/messages/en.json
Weegy: `markdownTable.ariaCloseBackdrop` is never used by `MarkdownTable.tsx` (the modal is a full-screen `fixed inset-0` panel with no separate backdrop element). The same unused key is added to `de.json`. Either drop it from both locales or add a real backdrop element that uses it.

Weegy (changes requested):
**Requesting changes.** Thanks for tackling the wide/tall table problem. The portal-rendered fullscreen modal and the SSR/hydration deferral are well done, and hoisting the components map to module scope is the right call for streaming.

A few items to address before merge. The first is a behavioral regression on another surface; the rest are correctness/accessibility/cleanup. Details are inline.

1. **Regression (Medium):** the change to the global `.md-view table` rule reaches `BuilderMarkdown.tsx`, which renders GFM tables into `.md-view` without the new scroll wrapper. See inline comment on `globals.css`.
2. **Sticky header border (Medium):** `position: sticky` + `border-collapse: collapse` tends to drop the header border on scroll. Please verify in-browser.
3. **Maximize button overlap (Medium):** the button overlays the top-right header cell and shows on every table.
4. **Accessibility (Medium):** the scroll container is not keyboard-focusable, and the modal has no focus trap / body scroll lock / focus restoration.
5. **Dead i18n key (Low):** `markdownTable.ariaCloseBackdrop` is unused in both locales.

Happy to pair on the CSS scoping fix.

web-ui/app/globals.css
```diff
@@ -480,18 +480,67 @@ body {
...
 .md-view th {
   @apply font-semibold;
   background: var(--bg-soft);
 }
+/* `border-collapse: separate` keeps cell borders on the cell, so a sticky
+   header keeps its divider while scrolling. We still render the bottom rule
+   as an inset shadow (instead of `border-bottom`) so it's painted by the
+   sticky cell itself — guaranteed visible regardless of how the row above
+   resolves the shared edge. */
+.md-view thead th {
```
ConnysCode: **Please scope these rules to `.md-table-wrap`, not `.md-view`.** `BuilderMarkdown.tsx` renders tables into `.md-view` via its own `<ReactMarkdown>` (it overrides only the `ol` slot), so it never gets the `.md-table-wrap` wrapper — yet this `position: sticky` rule still applies to its `<thead>`. With no local scroll container, the header pins to the builder chat pane instead (`BuilderChatPane.tsx:544`, `SimpleIntakePane.tsx:241`, `PreviewChatPane.tsx:506` are all `overflow-y-auto`). Please change the selector to `.md-table-wrap thead th`.

Same root cause for `.md-view table` just above (lines 494-497): the dropped `my-3` / `w-full` and `border-collapse: separate` also reach BuilderMarkdown (it loses its vertical margin). Please move those under `.md-table-wrap` too so non-wrapped surfaces stay at their previous default.

web-ui/app/_components/MarkdownTable.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function MarkdownTable({
+  children,
+  className,
+}: TableProps): React.ReactElement {
+  const t = useTranslations('markdownTable');
+  return (
+    <div
+      className="md-table-wrap"
+      tabIndex={0}
+      role="region"
```
ConnysCode: **Prefer `role="group"` over `role="region"`.** `region` is a landmark — one per table means any view with several tables fills the landmark/rotor menu with identically-named "Scrollable table" entries. `role="group"` keeps the focusable, labelled scroll container without injecting a landmark. The `tabIndex` + localized `aria-label` are otherwise good.

web-ui/app/_components/Markdown.tsx
```diff
@@ -117,9 +118,26 @@ export function Markdown({
...
+      >
         {displayed}
       </ReactMarkdown>
     </div>
   );
 }
+
+/**
+ * GFM table renderer override. Wraps every table in a scroll-container with
+ * a "full view" toolbar — see {@link MarkdownTable}. Hoisted to module scope
```
ConnysCode: **Stale comment (NIT).** There's no "full view" toolbar anymore — the modal was dropped this revision and `MarkdownTable` is just the scroll wrapper. Please drop "with a 'full view' toolbar" so the JSDoc matches the code.

ConnysCode (changes requested):
**Review: Request changes**

One item left before merge; everything else from my previous review is addressed (those threads resolved).

### Must change before merge

🟠 **MAJOR — the global table rules leak onto `BuilderMarkdown`.** The `.md-table-wrap` scroll wrapper is added only by the `table` override in `Markdown.tsx`. The rewritten `.md-view thead th { position: sticky }`, `.md-view table { border-collapse: separate }` and the dropped `my-3` / `w-full` are global, not wrapper-scoped. `BuilderMarkdown.tsx` renders into `.md-view` via its own `<ReactMarkdown>` (it overrides only the `ol` slot), so its tables get the new rules with **no** scroll container — the `<thead>` pins to the builder chat pane (`BuilderChatPane.tsx:544`, `SimpleIntakePane.tsx:241`, `PreviewChatPane.tsx:506`, all `overflow-y-auto`) and the table loses its margin. This is the cross-surface bleed `BuilderMarkdown`'s own docstring warns against. Scope the rules to `.md-table-wrap` (`.md-table-wrap thead th { position: sticky … }`, and move `border-collapse` / `my-3` under `.md-table-wrap`) — or correct the description and own the builder change. Inline detail below.

- Test migrated to `renderWithIntl` — CI blocker gone. Verified on `4b34ab0`: `vitest` 4/4 pass, `tsc --noEmit` clean, `eslint` clean.
- Modal dropped → focus-trap / `inert` / motion-token / print-button / button-overlap threads moot (resolved).
- Sticky-header border → `border-collapse: separate` + inset-shadow divider, as requested (resolved).
- Scroll region → `tabIndex={0}` + `role` + `aria-label` added (resolved).
- Dead `ariaCloseBackdrop` key removed (resolved).

| Claim | Reality |
|---|---|
| "Wrapper scoped to .md-table-wrap — BuilderMarkdown and other .md-view surfaces stay at their previous default" | ❌ Only `min-width`/`width` are wrapper-scoped. `.md-view thead th` (sticky), `.md-view table` (border-collapse, dropped `my-3`/`w-full`) and the `.md-view th/td` rewrite stay global, so BuilderMarkdown tables do change. |
| "One new i18n key" | ✅ Only `markdownTable.scrollRegionLabel`, present in de+en, parity passes. |
| "No schema / API / env" | ✅ Accurate. |
| Test plan (unticked) | lint / tsc / vitest all pass on `4b34ab0` (verified) — tick them. |

### Minimum to merge

1. Scope the sticky-header rule (and `border-collapse` / `my-3`) to `.md-table-wrap` so non-wrapped `.md-view` surfaces (BuilderMarkdown) stay unchanged — or correct the description and own the builder change.

Non-blocking: `role="region"` → `role="group"` (landmark spam), and add a small `MarkdownTable` test (25-file vitest convention). Inline.

- `Markdown.tsx:134` — stale JSDoc "full view toolbar"; no toolbar anymore (modal dropped).
- `MarkdownTable.tsx:27` — `tabIndex={0}` is unconditional, so non-scrolling tables enter the tab order. Defensible per WCAG 2.1.1; noting only.
- Border model (separate + inset-shadow) and the `@media print` fallback are correct — verified.
</details>

*(Reviewed against head `4b34ab0`.)*

sneumannb5:
changes implemented, see latest commit

ConnysCode (dismissed):
**Review: Approve**

All blocking and major findings from my two prior reviews are addressed and verified on head `e3ba05a`. Approving.

- **MAJOR — the `.md-view` → `BuilderMarkdown` CSS leak is fixed.** The sticky-header, `border-collapse: separate`, `min-width: max-content` and `width`/`margin` rules moved from `.md-view` to `.md-table-wrap` (`globals.css:516-551`). `BuilderMarkdown` renders into `.md-view` *without* the wrapper (`BuilderMarkdown.tsx:27-41` — `ol`-only override), so its tables now fall back to the classic `.md-view table` collapse model: no sticky `<thead>` pinning to the builder pane, margin intact. The cross-surface bleed its own docstring warns about is gone.
- `role="region"` → `role="group"` on the scroll wrapper (`MarkdownTable.tsx:28`) — no landmark spam.
- Stale "full view toolbar" JSDoc dropped (`Markdown.tsx:134`).
- Added `MarkdownTable.test.tsx` (wrapper contract / focusable group + i18n aria-label / className forwarding).
- Test-plan checkboxes ticked.

Verified on a fresh clone at `e3ba05a` (`web-ui/`):
```
vitest run             → 7/7 pass (Markdown 4 + MarkdownTable 3)
tsc --noEmit           → clean
eslint (changed files) → clean
i18n-validate          → OK, 1296 keys (scrollRegionLabel present en+de)
```
Also ruled out a specificity trap: `.lume-prose .md-view table` (`globals.css:106-113`) outranks `.md-table-wrap table`, but it only sets `font-family`/`font-size`, so it doesn't override the wrapper's border/sticky model.

| Claim | Reality |
|---|---|
| "Wrapper scoped to .md-table-wrap — BuilderMarkdown and other .md-view surfaces stay at their previous default" | Substantially accurate now. Sole residual app-wide change: `overflow-wrap: break-word` on `.md-view th/td` (`globals.css:502`) — cosmetic, prevents long-word overflow, not a regression. |
| "One new i18n key" | ✅ `scrollRegionLabel`, en+de, parity passes. |
| "No schema / API / env" | ✅ Accurate. |
| Test plan (now ticked) | ✅ lint / tsc / vitest verified green on `e3ba05a`. |

### Non-blocking (optional)
- `MarkdownTable.tsx:27` — `tabIndex={0}` is still unconditional, so a table that doesn't actually overflow still takes a tab stop. Defensible under WCAG 2.1.1, and you chose this last round — noting only.
- The visual cases (sticky divider holding on scroll, no border doubling on the wrapper edge) rest on your manual test plan; I can't re-run those headlessly.

*(Reviewed against head `e3ba05a`.)*

Weegy (approved):

PR #392, opened 2026-06-30
feat(ui-prefs): persist Lume palette/appearance server-side per user …

### LINKED ISSUE #287: Lume: palette preference should persist server-side per user/context (visual-spec §2.5.4)
§2.5.4 binds the accent palette per `memory://ui-prefs/<tenantId>/<userId>/<contextKey>/accent`. The operator UI (PR #284) persists the palette + appearance choice in `localStorage` — per browser, not per user, no multi-device sync.

Follow-up: move the preference to a server-side ui-prefs store (per user; contextKey can collapse to a single operator context for now) and hydrate it into the pre-paint bootstrap (e.g. via a cookie mirror to keep the no-FOUC property).

Out of scope in #282 by design ("at most a plain setting later"); this issue is that "later".

sneumannb5:
## What
  Move Lume palette/appearance from per-browser localStorage to a per-user server store (Closes #287). New `/api/v1/ui-prefs`
  GET/PUT backed by MemoryStore; RSC layout reads a mirrored cookie for no-FOUC pre-paint.

  ## Why
  localStorage was per-device — no multi-device sync. §2.5.4 wants the choice per user.

  ## Test plan
  - [x] `npm run test` middleware — uiPrefs router 9/9
  - [x] `npm run test` web-ui — ThemeControls
  - [x] manual: change palette → reload → new device sees it

  ## Risk / blast radius
  - No schema/migration (reuses MemoryStore), no new env var.
  - Scratch purge protects `ui-prefs`; full `reseed` still clears it.

### DIFF
--- a/middleware/src/routes/uiPrefs.ts
+++ b/middleware/src/routes/uiPrefs.ts
@@ -1,0 +1,141 @@
+import { Router } from 'express';
+import type { Request, Response } from 'express';
+import { z } from 'zod';
+
+import type { MemoryStore } from '@omadia/plugin-api';
+
+/**
+ * Per-user UI preferences — the server-side home for the Lume palette +
+ * appearance choice (issue #287, visual-spec §2.5.4).
+ *
+ * §2.5.4 binds the accent palette per
+ * `memory://ui-prefs/<tenantId>/<userId>/<contextKey>/accent`. The first Lume
+ * integration (#284) parked the palette + appearance choice in localStorage —
+ * per browser, no multi-device sync. This router moves it into the MemoryStore
+ * so the choice follows the user across devices. The operator UI is a single
+ * context today, so `tenantId` and `contextKey` collapse to the constants
+ * `default` / `operator` (per the ticket: "contextKey can collapse to a single
+ * operator context for now").
+ *
+ * The web-ui mirrors the value into a non-secret cookie so the pre-paint
+ * bootstrap can set `data-palette`/`data-theme` on <html> with no FOUC; this
+ * store stays the source of truth that seeds that cookie on a fresh device.
+ *
+ * Mounted under `/api/v1/ui-prefs`, gated by `requireAuth`.
+ *   GET /  → current prefs ({} when none stored yet)
+ *   PUT /  → upsert { palette?, appearance? }
+ */
+
+const PALETTES = ['lagoon', 'petrol', 'atelier'] as const;
+const APPEARANCES = ['system', 'light', 'dark'] as const;
+
+const UiPrefsSchema = z
+  .object({
+    palette: z.enum(PALETTES).optional(),
+    appearance: z.enum(APPEARANCES).optional(),
+  })
+  .strict();
+
+export type UiPrefs = z.infer<typeof UiPrefsSchema>;
+
+export interface UiPrefsRouterDeps {
+  store: MemoryStore;
+  log?: (msg: string) => void;
+}
+
+function errMsg(err: unknown): string {
+  return err instanceof Error ? err.message : String(err);
+}
+
+/** Map a session user id onto a path-safe MemoryStore segment. `omadia_user_id`
+ *  is a KG cluster id, but escaping keeps the path valid regardless of the id
+ *  shape the provider mints. The escaping is INJECTIVE — every char outside
+ *  `[A-Za-z0-9-]`, including `.` and `_` themselves, becomes `_<4-hex code
+ *  unit>` — so (a) two distinct ids can never collapse onto the same path
+ *  (which would leak one user's prefs to another), and (b) no segment can ever
+ *  be `.` or `..`, so a `..` id can't traverse out of the per-user directory. */
+function safeSegment(id: string): string {
+  return id.replace(
+    /[^A-Za-z0-9-]/g,
+    (c) => `_${c.charCodeAt(0).toString(16).padStart(4, '0')}`,
+  );
+}
+
+function prefsPath(userId: string): string {
+  return `/memories/ui-prefs/default/${safeSegment(userId)}/operator.json`;
+}
+
+/** Read + validate the stored prefs, or `{}` when absent. A stored value that
+ *  is corrupt (non-JSON), pre-dates a schema change, or got hand-edited is
+ *  treated as unset rather than surfaced as an error — the client falls back
+ *  to its defaults, and crucially a corrupt file does not brick the read-merge
+ *  PUT path (which would otherwise 500 on every write, leaving no way to
+ *  overwrite the bad value). Real IO errors still propagate to the caller. */
+async function readPrefs(store: MemoryStore, path: string): Promise<UiPrefs> {
+  if (!(await store.fileExists(path))) return {};
+  const raw = await store.readFile(path);
+  try {
+    const parsed = UiPrefsSchema.safeParse(JSON.parse(raw));
+    return parsed.success ? parsed.data : {};
+  } catch {
+    return {};
+  }
+}
+
+function requireSessionUserId(req: Request, res: Response): string | null {
+  const id = req.session?.omadia_user_id;
+  if (!id) {
+    res.status(401).json({ code: 'auth.required', message: 'login required' });
+    return null;
+  }
+  return id;
+}
+
+export function createUiPrefsRouter(deps: UiPrefsRouterDeps): Router {
+  const router = Router();
+  const log = deps.log ?? ((m) => console.log(m));
+
+  router.get('/', async (req: Request, res: Response): Promise<void> => {
+    const userId = requireSessionUserId(req, res);
+    if (!userId) return;
+    try {
+      res.json(await readPrefs(deps.store, prefsPath(userId)));
+    } catch (err) {
+      // Log the detail; return a generic message so an internal store error
+      // (paths, driver internals) is not echoed back to the client.
+      log(`[ui-prefs/route] GET / failed: ${errMsg(err)}`);
+      res
+        .status(500)
+        .json({ code: 'ui_prefs.read_failed', message: 'failed to read ui prefs' });
+    }
+  });
+
+  router.put('/', async (req: Request, res: Response): Promise<void> => {
+    const userId = requireSessionUserId(req, res);
+    if (!userId) return;
+    const parsed = UiPrefsSchema.safeParse(req.body);
+    if (!parsed.success) {
+      res.status(400).json({
+        code: 'ui_prefs.invalid_request',
+        issues: parsed.error.issues,
+      });
+      return;
+    }
+    try {
+      // MERGE, not replace: the fields are individually optional, so a PUT of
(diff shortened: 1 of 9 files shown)


middleware/src/routes/uiPrefs.ts
```diff
@@ -0,0 +1,141 @@
...
+  try {
+    const parsed = UiPrefsSchema.safeParse(JSON.parse(raw));
+    return parsed.success ? parsed.data : {};
+  } catch {
+    return {};
+  }
+}
+
+function requireSessionUserId(req: Request, res: Response): string | null {
+  const id = req.session?.omadia_user_id;
```
ConnysCode: 🟠 **MAJOR — 401 here bounces a validly-authenticated user to `/login`.** `omadia_user_id` is optional on a live session: `signSession` only sets it when `resolveChannelIdentity` resolves, which returns `undefined` in the eventual-consistency window of a new user's first OIDC login (`index.ts:2249-2254`) and on any KG error (caught at `auth.ts:608`). The client reads this 401 as an expired session — `getUiPrefs` (via `getJson`) and `putUiPrefs` both call `maybeNavigateToLogin(401)` — so the user is redirected to `/login` on page load and on every palette change, with re-login looping while the KG stays unresolved.

This router is mounted behind `requireAuth`, which guarantees `req.session.sub` (a required claim). Please key the store on the identity that's always present:
```ts
const id = req.session?.omadia_user_id ?? req.session?.sub;
```
Then the 401 only fires for a genuinely session-less request (the router's own unit-test path), and the feature works for every authenticated user.

web-ui/app/_components/ThemeControls.tsx
```diff
@@ -64,32 +76,69 @@ export function ThemeControls(): React.ReactElement {
...
+
+  // Hydrate from the server store on mount: the cookie/SSR value may be stale
+  // (or absent on a fresh device). Apply the stored choice to <html> — the
+  // MutationObserver re-renders the selects — and refresh the cookie so the
+  // next pre-paint on this device matches. No PUT here: the store is already
+  // the source. Stays silent when logged out / offline (cookie value holds),
+  // or when the user has already made a choice this session (their PUT wins).
+  useEffect(() => {
+    let cancelled = false;
+    void getUiPrefs()
```
ConnysCode, line 93: 🟠 **MAJOR (client half of the same issue) — this mount fetch can redirect the page to `/login`.** `getUiPrefs()` goes through `getJson`, which calls `maybeNavigateToLogin(res.status)` on any non-OK response *before* rejecting — so the `.catch()` below cannot stop it. When the session lacks `omadia_user_id` the router 401s (see my comment on `uiPrefs.ts`), and a user who just logged in is bounced straight back to `/login` on first paint. Fixing the router to fall back to `sub` removes the trigger; if you also want defence-in-depth, please have these best-effort prefs calls bypass the login-bounce (e.g. a `getJson`/`putUiPrefs` option that skips `maybeNavigateToLogin`) so a passive preference fetch can never hijack navigation.

web-ui/app/_components/ThemeControls.tsx
```diff
@@ -2,37 +2,49 @@
...
-  return v === 'system' || v === 'light' || v === 'dark';
+/** Mirror the choice into the non-secret cookie the RSC layout reads pre-paint.
+ *  Lax + path=/ so it rides every same-site navigation; not httpOnly because
+ *  it carries no secret and the client both writes and (via SSR) consumes it.
+ *  `Secure` over HTTPS so it is never sent on a downgraded plain-HTTP request;
+ *  omitted on http (localhost dev) where browsers reject Secure cookies. */
+function writeUiPrefsCookie(palette: PaletteName, appearance: Appearance): void {
+  const value = encodeURIComponent(JSON.stringify({ palette, appearance }));
+  const secure = window.location.protocol === 'https:' ? ';secure' : '';
+  document.cookie = `${UI_PREFS_COOKIE}=${value};path=/;max-age=${COOKIE_MAX_AGE};samesite=lax${secure}`;
```
ConnysCode, line 47: 🟡 **MINOR — this cookie outlives the session.** `max-age` is 1 year and `POST /logout` clears only `omadia_session`, not `omadia-ui-prefs` (`auth.ts:296`). On a shared browser the next user's first server paint renders the *previous* user's palette/theme until the mount `getUiPrefs()` corrects it. It carries no secret, but it's a visible cross-user flash. Please clear `omadia-ui-prefs` in the logout handler alongside the session cookie.

ConnysCode (changes requested):
**Review: Request changes**

One MAJOR before merge: a 401 from this endpoint bounces a *validly-authenticated* user to `/login`. The router design, the storage/escaping, the purge protection and the tests are otherwise sound (verified below).

### Must change before merge

🟠 **MAJOR — a 401 from `ui-prefs` redirects a logged-in user to `/login`.** `requireSessionUserId` returns 401 whenever the session has no `omadia_user_id` ([uiPrefs.ts:86](https://github.com/byte5ai/omadia/blob/3292d26/middleware/src/routes/uiPrefs.ts#L86)). That claim is **optional** on a valid session — `signSession` only sets it when `resolveChannelIdentity` succeeds, and that resolver returns `undefined` in the documented eventual-consistency window of a brand-new user's first OIDC login (`index.ts:2249-2254`: "Returns undefined if the users-row was just created in the same request"), and `undefined` again on any KG hiccup (the `catch` at `auth.ts:608` swallows it). The client treats this 401 as a dead session: the mount `getUiPrefs()` runs through `getJson` → `maybeNavigateToLogin(401)` and the palette-change `putUiPrefs()` does the same — so such a user is **redirected to `/login` on page load** (and again on every palette change), even though their session is live. Re-login doesn't help while the KG is unresolved → a loop. `requireAuth` already guarantees `req.session.sub` (a required claim). Please key the store on `req.session.omadia_user_id ?? req.session.sub` so a live session never 401s here; the router's 401 then only fires for a genuinely session-less request (its own unit-test path).

### Should fix

🟡 **MINOR — the prefs cookie is not cleared on logout.** `POST /logout` clears `omadia_session` but not `omadia-ui-prefs` (`auth.ts:296`; the cookie has a 1-year max-age). On a shared browser the next user's first paint shows the previous user's palette until the mount `getUiPrefs()` corrects it. Non-secret, but a visible cross-user flash. Please clear `omadia-ui-prefs` in the logout handler too.

| Claim | Reality |
|---|---|
| "uiPrefs router 9/9" | ✅ Confirmed — 9 `it()` blocks in `uiPrefsRoute.test.ts`, exercising round-trip, partial merge, injective-id isolation (`a b` vs `a_b`), `..` containment, corrupt-file recovery, and the 401 guard. |
| "Scratch purge protects `ui-prefs`; full reseed still clears it" | ✅ Confirmed — `topLevelName('/memories/ui-prefs/…')` → `ui-prefs`, which is in `PROTECTED_SEED_ENTRIES`; skipped under `axis:'all'` unless `reseed`. |
| "per-user server store … the choice follows the user across devices" (durable) | ⚠️ Only with the Postgres store. The **default** `memoryStore` is the RAM-backed `InMemoryMemoryStore` (`harness-memory/plugin.ts:51` — "DB-less, RAM-backed … default provider when no Postgres-backed store is installed"); without `@omadia/memory-postgres` the prefs are lost on every middleware restart. Worth stating in the PR/issue so the durability expectation matches the deployment. |
| "No schema/migration … no new env var" | ✅ Accurate. |

### Minimum to merge
1. Fall back to `sub` for the storage key (or otherwise stop a missing-claim 401 from hijacking navigation) so first-login / KG-degraded users aren't bounced to `/login`.
2. Clear the `omadia-ui-prefs` cookie on logout.

- The palette/appearance enums are duplicated in `middleware/src/routes/uiPrefs.ts` and `web-ui/app/_lib/uiPrefs.ts` (unavoidable across the two packages). Consider a cross-reference comment in each so a spec addition (a 4th palette) isn't silently applied to only one side — the server would 400 the new value the client offers.
- `readPrefs` swallows a corrupt stored value as `{}` (good, and tested), but a genuine parse error is indistinguishable from "unset" in logs. Optional: `log()` the corrupt-discard so a recurring bad write is visible.
</details>

*(Reviewed against head `3292d26`.)*

ConnysCode (approved):
**Review: Approve**

Both items from my previous review are addressed and verified against the code at head `20f207d`. No blockers remain — the two NITs below are optional.

- **MAJOR (401 bounce) fixed.** `requireSessionUserId` now keys on `req.session?.omadia_user_id ?? req.session?.sub` ([uiPrefs.ts:93](https://github.com/byte5ai/omadia/blob/20f207d/middleware/src/routes/uiPrefs.ts#L93)). `sub` is a required claim (`SessionClaims.sub` is non-optional and `verifySession` throws when it is empty — `sessionJwt.ts:96`), and `requireAuth` runs before the router, so a live session can no longer 401 here. The router's 401 now only fires for a genuinely session-less request — exactly the unit-test path.
- **MINOR (prefs cookie on logout) fixed.** `POST /logout` now `res.clearCookie('omadia-ui-prefs', { path: '/' })` alongside the session cookie ([auth.ts:298](https://github.com/byte5ai/omadia/blob/20f207d/middleware/src/routes/auth.ts#L298)); name + `path=/` match the cookie `ThemeControls` writes, so the delete takes effect and the cross-user first-paint flash is gone.
- **Coverage grew with the fix.** A dedicated `sub`-fallback test was added and the session-less test now drives a null session (`uiPrefsRoute.test.ts:225`, `:238`) — 10 `it()` blocks total.

| Claim | Reality |
|---|---|
| "uiPrefs router 9/9" | Now **10** `it()` blocks — the fix added the `sub`-fallback case. Cosmetic; please bump the number if you touch the description. |
| "Scratch purge protects `ui-prefs`; full reseed still clears it" | ✅ Still accurate — `'ui-prefs'` is in `PROTECTED_SEED_ENTRIES` (`memoryPurge.ts:43`). |
| "No schema/migration … no new env var" | ✅ Accurate. |

### Note (non-blocking, verification not executed)
I did not re-run the suite this pass — a full monorepo install is heavy and the delta is a self-contained new test plus a null-session tweak. My previous pass confirmed the suite green; the changed paths are verified by reading. If CI is green on this head, treat the test claim as confirmed.

- `middleware/src/routes/uiPrefs.ts:93` — the `sub` fallback introduces a one-time key change: prefs a user sets during the first-login / KG-unresolved window are stored under `sub`, then read under `omadia_user_id` once the KG resolves and a new session is minted — so that early choice is silently orphaned and the user sees defaults again. Rare and self-healing (they just re-pick), and strictly better than the old 401 loop. Optional: a short comment noting the key can migrate, or reading `sub` as a fallback on GET too.
- The palette/appearance enums are still duplicated across `middleware/src/routes/uiPrefs.ts:29-30` and `web-ui/app/_lib/uiPrefs.ts:6-9` (unavoidable across the two packages). Optional: a cross-reference comment so a 4th palette isn't added to only one side (the server would then 400 a value the client offers).
</details>

*(Reviewed against head `20f207d`.)*

Issue #92, opened 2026-05-18
web-ui: bump eslint 9 → 10 once eslint-plugin-react v8 stable releases


Weegy:
## Context

Tracked from the eslint 9 → 10 migration sweep (PR #90 covered middleware-only; #81 is the Dependabot tracker that will reopen).

`web-ui` cannot upgrade to eslint 10 yet because `eslint-config-next@16` brings in `eslint-plugin-react@^7.37.5` (vendored under its own `node_modules`), whose stable releases declare:

```json
"peerDependencies": { "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7" }
```

eslint 10 falls outside that range and trips a runtime crash inside `eslint-plugin-react/lib/util/version.js`:

```
TypeError: Error while loading rule 'react/display-name':
  contextOrFilename.getFilename is not a function
```

The only on-npm version that supports eslint 10 is `eslint-plugin-react@7.8.0-rc.0` (pre-release — not used).

## What to do when this unblocks

1. Watch <https://github.com/jsx-eslint/eslint-plugin-react/releases> for a v8 stable.
2. When released, bump `eslint-plugin-react` via npm overrides in `web-ui/package.json` OR wait for `eslint-config-next` to ship a release that pins it transitively.
3. Bump `web-ui/package.json` → `"eslint": "^10.x"`, `npm install`, expect a clean lint (lock-update only).
4. Close this issue + #81.

## Acceptance

- `npm run lint` in `web-ui/` passes under eslint 10 without crashing.
- No re-introduction of `--legacy-peer-deps`.

Weegy:
## Triage
**Priority:** P3 · **Category:** maintenance · **Effort:** S · **Status:** Blocked on upstream (eslint-plugin-react v8)

[...]

PR #469, opened 2026-07-09
chore(scripts): add interactive issue-loop workflows (cluster + implement)

Weegy:
## What

Two Claude Code Workflow-tool scripts that turn the open backlog into a guided **cluster → approve → implement** loop:

| File | Stage | Writes to GitHub? |
|---|---|---|
| `scripts/issue-cluster.workflow.mjs` | read all open issues → enrich → cluster → rank | no |
| `scripts/issue-implement.workflow.mjs` | branch → code → adversarial review | no |

They sit alongside the existing `scripts/issue-triage.workflow.mjs` and follow its contract exactly: **the workflow computes, the caller writes.**

## Why two scripts and not one

A Workflow script cannot call `AskUserQuestion` — only the main Claude loop can. The requested flow has two human gates in the middle ("which group do we start with?", then "implement, skip, or defer?" per issue). Those gates force the split. The full choreography, including the run-state file format that survives compaction, is documented in the header of `issue-cluster.workflow.mjs`.

## The load-bearing idea: `cohesion`

Clustering is irreducibly global — an agent that only sees its own issue cannot decide which other issues it belongs with. So the map stage runs one code-grounded investigator per issue in parallel, and a single reduce agent groups the compact descriptors. The grouping signal is **overlapping `touchedFiles`**, not title similarity.

Each cluster is then tagged:

- **`dependent`** — the issues share files or symbols. One agent implements them in priority order on **one branch**, producing **one pull request** closing all of them.
- **`independent`** — disjoint files. One worktree-isolated agent and one pull request per issue, in parallel.

This distinction is the difference between working and self-destructing. "Group A: 3 auth issues (related)" means those issues mutate the same middleware file. Fanning them out to three parallel agents yields three pull requests that each revert the other two. When file overlap is real, the cluster serialises; when in doubt, the reduce agent is instructed to tag `dependent`, because serialising costs time and colliding costs the whole run.

Intra-cluster ranking (`impact / effort`) is deterministic JavaScript, not an agent call, so the order does not drift between runs.

## Safety model

- Neither script pushes, opens a pull request, merges, or enables auto-merge. Branch refs created inside agent worktrees survive to the caller, which performs the single remote write — downstream of the human's per-issue greenlight.
- Every branch must clear an **adversarial reviewer** (secrets, scope creep, forbidden paths under `.github/`, correctness, and whether the implementer's `verifyLog` shows a real passing gate or a skipped one) before it is reported as `prReady`. A false green costs a bad merge; a false red costs one more look.
- Slugs are sanitized **in JavaScript**, not by prose rule. They are LLM output derived from issue titles that strangers write, and they end up inside a branch name an agent interpolates into `git checkout -b`. Prose ground rules do not stop shell metacharacters.
- Issue numbers and `baseSha` are validated; the implement set is hard-capped at 8; duplicate placements from the reduce agent are de-duplicated (first cluster wins).
- REST-only `gh api`; the scripts contain no `gh issue` / `gh pr` / `gh search` / GraphQL call.

## Verification

- Both files parse (`node --check`, wrapped to allow the runtime's top-level `return`), contain zero null bytes, and stay under the 500-line limit (352 / 342).
- Every `gh issue` / `gh pr` string in either file is part of a *prohibition*, confirmed by inspection.
- The de-duplication fix was exercised against a synthetic reduce-agent output that places `#12` in two clusters and hallucinates a `#999`: `#12` lands in exactly one cluster, `#999` is dropped, and the cluster that shrinks to a single member is correctly downgraded from `dependent` to `independent`.
- Designed via a three-way architecture panel and reviewed by five independent adversarial lenses (runtime-API, git/worktree, blast-radius, logic, fidelity); each finding was then independently verified. Seven raw findings, six refuted, one confirmed and fixed (the duplicate-placement bug above).

**Not yet exercised end-to-end against the live backlog.** The scripts are static, reviewed, and unit-checked; the first real run should use a small `maxIssues` on an `independent` cluster.


Weegy:
## First live run — one real defect found and fixed

Ran `issue-cluster.workflow.mjs` against all 28 open issues of this repo (read-only; no branches, no GitHub writes). 29 agents, 0 errors, ~6 minutes.

Mechanically it worked: 28/28 issues placed, 0 duplicates, deterministic ranking. But the cohesion output was **7 of 7 clusters `dependent`, 0 `independent`** — meaning the fan-out path could never trigger and every cluster would have been chained onto one oversized branch.

The reduce agent was citing `web-ui/messages/en.json`, `middleware/src/index.ts` and `README.md` as evidence of coupling. Those are **hub files**: registration points, i18n catalogs, config modules, docs. Every feature appends to them, they merge cleanly, and they say nothing about whether two issues collide. Together with the prompt's "when in doubt, tag dependent" tie-break, they dragged every cluster to `dependent`.

A file-overlap histogram over the run confirmed 11 such files (touched by ≥ 3 of 28 issues):

    4x  middleware/src/index.ts
    4x  middleware/packages/plugin-api/src/pluginContext.ts
    3x  web-ui/messages/{en,de}.json
    3x  README.md
    3x  middleware/src/config.ts        (+ 5 more)

`findHubFiles()` now computes these deterministically (`>= max(3, 10% of issues)`) and strips them from the `touchedFiles` line the reduce agent judges cohesion on. They are still shown per issue as `alsoTouchesHubFiles` so the agent understands what an issue does. The cohesion rules now also state the cost of a **false `dependent`** — unrelated issues chained onto one branch, one bad issue blocking the rest — rather than treating `dependent` as the safe default, and require a named non-hub path or symbol in `cohesionEvidence`.

### Before → after, same 28 issues

Re-ran with `resumeFromRunId`, so the 28 enrich agents replayed from cache and only the reduce stage re-ran (56k tokens instead of 1.47M).

| | before | after |
|---|---|---|
| clusters | 7 | 9 |
| `dependent` | **7** | 7 |
| `independent` | **0** | **2** |
| largest cluster | 5 issues | 2 issues |
| unclustered | 2 | 8 |

Every surviving `dependent` tag is now backed by real evidence: migration-number collisions (`0006_webhook_endpoints.sql` vs `0006_ephemeral_workflows.sql`; `0029_datasets.sql` vs `0029_embedding_model_metadata.sql`), a shared `web-ui/eslint.config.mjs`, `harness-verifier/src/verifierPipeline.ts`, `harness-plugin-privacy-guard/src/service.ts`, and root `docker-compose.yaml`.

The old five-issue cluster that would have chained #403 (ready-to-implement) together with #92 (blocked-external) and #286 (needs-decision) onto a single branch is gone. #403 now stands alone.

### Still not exercised

`issue-implement.workflow.mjs` has **not** been run. Nothing has branched, committed, or pushed. The first implement run should take a single `independent` issue with `maxIssues: 1`.

Weegy:
## First end-to-end run: the loop produced PR #472

Ran the full loop against the live backlog. The cluster stage grouped all 28 open issues; a human picked issue #403 from the gate, approved it, and the implement stage ran. Result: **PR #472**, approved by both reviewers.

It took six implement→review rounds, and the interesting part is *who* found *what*.

| Round | Reviewer | Finding |
|---|---|---|
| 1 | Claude | The fix missed `SimpleIntakePane`, which `Workspace.tsx` shows is the **default** builder view |
| 2 | codex | A fourth surface (the stream toast) still rendered the raw error; missing changelog entry |
| 3 | codex | The toast reported `done` for a **failed** turn — a state bug, not a string bug |
| 4 | codex | The extractor destroyed clean application messages containing braces; commit subjects over 70 chars |
| 5 | codex | Status-less rate-limit JSON leaked raw to users — a flaw in the spec a human had written |
| 6 | — | Both reviewers approve |

**The Claude reviewer set `prReady: true` in rounds 2, 3, 4 and 5.** Without the cross-vendor stage, five distinct defects would have shipped as a pull request — two of them in the exact error class the issue was filed about. The reviewers were also complementary rather than redundant: the Claude reviewer caught that an implement agent under-reported its own diff (3 files claimed, 8 actual), which codex missed.

Round 5 is the one worth dwelling on. Codex rejected a spec **I** had written, not the agent's execution of it. I had told the agent to key the extractor's give-up on a leading HTTP status; codex pointed at `middleware/test/orchestrator/streamingRetry.test.ts`, which constructs a status-less `{"type":"error","error":{"type":"rate_limit_error"}}`, and showed it would reach the user as raw JSON. The correct discriminator turned out to be *does the whole string parse as a JSON object*, which also closed a latent edge the first reviewer had only footnoted.

### Changes to the workflow driven by these runs

- **Hub files poisoned the cohesion signal.** The first cluster run returned 7 of 7 clusters `dependent`, killing the fan-out path entirely, because the reduce agent cited `en.json`, `index.ts` and `README.md` as evidence of coupling. `findHubFiles()` now subtracts files touched by many unrelated issues. Result on the same 28 issues: 9 clusters, 7 dependent / 2 independent, largest cluster down from 5 issues to 2.
- **Reviewers now read `AGENTS.md` / `CONTRIBUTING.md`.** The Claude reviewer could not have found round 2's changelog finding, because its checklist never mentioned policy files.
- **Cross-vendor review** (`codex`, gpt-5.5, `--sandbox read-only`) runs after the first reviewer approves. `codexReview: 'auto'` skips it cleanly when the CLI is absent or unauthenticated — an optional reviewer must never break the pipeline. GOTCHA: `codex exec` blocks forever on an open stdin; every call needs `</dev/null`.
- **Fixup mode.** `resumeBranch` + `fixFindings` iterate on the rejected branch instead of cutting a new one. The fixup agent is told to argue back rather than comply silently.
- **A review criterion the loop could not satisfy.** Codex flagged over-long commit subjects while the fixup agent was forbidden to amend. That finding would have been re-reported forever. The agent may now reword its own commits, but only after `git ls-remote` proves the branch was never pushed.
- **Step 8: comment on the issue.** `Closes #<n>` only creates a timeline cross-reference. A reader of the issue learns a PR exists, not what was found. Now an explicit step.

### Still true

The workflow writes nothing to GitHub and never pushes. Every remote write in this run — the branch push, PR #472, the comment on #403 — was performed by the caller, after a human approved that specific issue.

PR #471, opened 2026-07-09
One-click deploy (Render + Fly.io) — and the four first-run fixes it surfaced

SecurID:
## Summary

Makes omadia one-click deployable to a visitor's **own** cloud account, and fixes everything the live smoke test surfaced along the way. The full flow was verified end to end on a real Fly.io org: deploy → login → readiness banner → key in → agents/channels/skills live → restart → routines live.

### 1. Deploy paths (new)

- **`render.yaml`** — backs a *Deploy to Render* button: middleware + admin UI from the published GHCR images, managed Postgres 17 with pgvector, persistent `/data` disk, `VAULT_KEY` generated by Render (`generateValue`), region `frankfurt`. Zero pasted secrets: the `/setup` wizard collects the LLM key at runtime. The web-ui reaches the middleware via `fromService`/`envVarKey: RENDER_EXTERNAL_URL`.
- **`fly/deploy.sh` + three machine configs** — Fly has no blueprint-style button, so the equivalent is one command. Provisions three apps: a **private-only `pgvector/pgvector:pg17`** (Fly's unmanaged postgres-flex ships without pgvector; Managed Postgres gates `vector` behind a dashboard toggle at a much higher price), middleware with a `/data` volume, and the web-ui — plus generated secrets. ~$10/month.
- README Deployment section documents both.

### 2. `MIDDLEWARE_URL` was frozen at image build time (fix)

Next evaluates `rewrites()` at build time; the published web-ui image had `http://middleware:8080` baked into its routes manifest, so **every browser call through `/bot-api` failed with `ENOTFOUND` on any non-compose host** while runtime `MIDDLEWARE_URL` was silently ignored. `/bot-api/*` and `/p/*` are now catch-all route handlers (`app/_lib/middlewareProxy.ts`) resolving the env **per request** — streaming both ways (SSE, ZIP uploads), multi-`Set-Cookie` preserved, redirects passed through. Build ARG removed from the Dockerfile, CI workflows, and the compose build overlay; `desktop-apps.yml` stops pretending to bake the kernel URL (the supervisor injects it at spawn — this also fixes the latent wrong-port proxy there). 6 regression tests.

### 3. Fresh installs now say what's missing (new)

Without an LLM key the orchestrator publishes nothing and every operator surface 503s — the UI showed raw `GET … failed: 503` strings. `RuntimeReadinessBanner` probes one operator route and, on the structured `multi_orchestrator_unavailable`, shows a card naming the cause with a link to `/admin/settings`; it clears itself once the key lands. EN/DE, 6 tests.

### 4. 17 broken `t.rich` placeholders (fix, regression from #447)

Messages used ICU argument syntax (`{toolName}`) where components pass chunk functions — next-intl splices the raw function into the output and React hard-crashes the page (RSC: "Functions cannot be passed to Client Components"). Affected: routines (crashed on a live instance), login no-providers, onboarding modal, builder preview, and four admin pages. All 12 keys converted to tag syntax in both locales; a sweep audit now finds zero. The locale-parity check couldn't catch this class (both locales consistently wrong) — a follow-up issue could add an argument-vs-tag cross-check against component usage.

### 5. `next dev` was broken on main (fix)

Tailwind v4 scans comments as class candidates; two test-file comments containing a literal bracket-class with `var(...)` compiled into invalid CSS and 500'd every page in dev (prod unaffected). Comments reworded.

## Verification

- Live Fly smoke instance: full first-run arc verified (proxy 500→200, login, banner, key save, routines after restart)
- web-ui: 198/198 tests, `tsc --noEmit`, ESLint, `i18n:check` clean
- Build paths: local dev ✓, local `next build` ✓, standalone server with runtime `MIDDLEWARE_URL` ✓ (proxy + routines verified against a stub middleware), Docker image ✓ (built 3× on Fly's remote builder, live-verified), compose ✓ (sets the env at runtime), desktop ✓ (supervisor sets it at runtime)
- Remaining gate: this PR's own CI run (workflows verified by parse + review only)

## Notes for review

- **Release ordering:** the deploy button/script pull `:latest` GHCR images — they only work for real visitors once a release containing the proxy fix is published. The omadia.ai `/docs` deploy section (separate PR in `omadia-homepage`) should merge after that release.
- Follow-up candidates (intentionally not in this PR): hot-mount routines on key save (only feature still needing a restart); i18n validator rule for the `t.rich` mismatch class.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SecurID:
Important Extension for future PR: routines is the only feature that can't hot-enable on key save (it registers at boot, hence a restart is needed), and the middleware log even hints at it. Making it late-mount like chat does would remove the last manual step from first-run

Weegy:
## Prep pass: merged `main` (90 commits), resolved 3 conflicts, 1 lint fix — now `mergeable: clean`, CI fully green

This branch had been sitting behind `main` since 2026-07-09 and was `DIRTY`. Rebased-by-merge and made merge-ready. **One resolution changed the meaning of section 4 of the description — please read that part before reviewing.**

### 1. `messages/{en,de}.json` — section 4 of this PR is superseded by #599, and the branch's version had become a regression

Both this branch and `main` fixed the *same* `t.rich` ICU-vs-tag bug, in **different shapes**:

| | shape |
|---|---|
| this branch | `<toolName></toolName>` (empty tag — content supplied by the component) |
| `main` (via #599) | `<toolName>manage_routine</toolName>` (content in the message) |

Every chunk handler on `main` renders `{chunks}` — `routines/page.tsx:152`, `login/page.tsx:176`, `RoutineTemplateEditor.tsx:223`, `admin/auth/page.tsx:85`, and the rest. So the content has to live in the message, and this branch's empty-tag form would render **empty `<code></code>` elements** — silently dropping the word. All 11 hunks per locale were resolved to `main`'s side.

The `runtimeReadinessBanner` keys this PR adds survive unchanged in both locales, and the catalog's only remaining empty tag pair (`<trashIcon></trashIcon>`) is legitimate — `CredentialsEditor.tsx:253` does `trashIcon: () => (...)` and ignores chunks.

**This PR's own proposed follow-up is confirmed necessary.** I mutated one key back to the empty-tag form and ran the full suite: **718/718 still passed**, and `i18n:check` passed too. `i18n-parity.test.ts` compares tag *names* across locales, and both shapes have identical tag names — it is structurally incapable of catching this class. Nothing guards it today.

### 2. `.github/workflows/publish-images.yml` — union

Kept this branch's `build_args: ''` for `web-ui` and `main`'s `dev-runner` / `dev-runner-daemon` matrix entries (epic #470 W1).

### 3. `RuntimeReadinessBanner.tsx` — lint fix (new commit)

`main`'s Lume button ratchet (#616) landed after this branch was written and rejects the raw `<button>` used for the dismiss action. Migrated to `<Button variant="secondary">`, following `ConfirmDialog`'s precedent for a cancel-style affordance. web-ui lint: 1 error → 0.

### Verification on the merged tree

- `i18n:check`, `tsc --noEmit`, ESLint (0 errors), **718/718 vitest** — all green locally, and CI is green on `256f9914` (9/9 real checks).
- **The headline fix re-verified directly:** built with a poisoned build-time value — `MIDDLEWARE_URL=http://build-time-poison:9999 npm run build` — and it appears **nowhere** in `.next`. `routes-manifest.json` retains only the static `/.well-known/omadia-ui` rewrite. The freeze is genuinely gone.
- Desktop path confirmed safe: `desktop/src/supervisor.ts:175` injects `MIDDLEWARE_URL` at spawn using the dynamically chosen kernel port, so dropping the build-time bake also fixes the latent wrong-port bug the description mentions. Compose sets it at runtime (`docker-compose.yaml:121`).
- Checked the new catch-all handlers for path escape: feeding `..` segments straight into `createMiddlewareProxy` does escape the `/api` prefix, but a real standalone Next server never delivers them — `/bot-api/%2e%2e/health` is normalized at the routing layer and never reaches the proxy. **No traversal issue.**

### Still open (unchanged from the description)

The release-ordering note stands: the deploy button and `fly/deploy.sh` pull `:latest` GHCR images, so they only work for real visitors once a release containing the proxy fix is published.

PR #476, opened 2026-07-09
feat(scripts): spec-driven wave workflows + W0 unit manifest (epic #470)

Weegy:
Prepares the subagent-driven implementation of epic #470 (integrated dev platform). No product code — three Workflow scripts and one manifest.

## What this adds

| File | Purpose |
|---|---|
| `scripts/wave-decompose.workflow.mjs` | wave spec → conflict-checked unit manifest |
| `scripts/wave-implement.workflow.mjs` | units → worktree-isolated branches, adversarially reviewed |
| `scripts/wave-verify.workflow.mjs` | merged tree → per-criterion pass/fail + security audit |
| `docs/dev-platform/w0-manifest.json` | the W0 decomposition: 11 units, acyclic, 8 security-sensitive |

## Design notes

**Cross-family review is the gate, not a formality.** The implementer (Engineer, Claude-family) and a same-family reviewer share a training corpus and therefore share blind spots. Forge (GPT-5.4) refutes every diff; Cato audits the security-sensitive units against the epic's structural invariants. A branch is `prReady` only when the implementer verified, the refuter could not overturn it, and — where it matters — the auditor found no invariant violation. This mirrors what #472 demonstrated: the Claude reviewer said `prReady: true` in four consecutive rounds and the cross-vendor reviewer rejected all four.

**Hub files are subtracted from the collision signal.** `index.ts`, `config.ts`, and `messages/*.json` are append-only registration points touched by nearly every unit. Left in the signal, every unit reads as dependent and the fan-out dies (the lesson `issue-cluster` learned the hard way). Instead the manifest assigns every hub edit to one `w0-wire` unit that `dependsOn` all others — a merge conflict becomes a dependency edge.

**`parallel()` vs `pipeline()` is deliberate.** Decomposition uses a barrier because the `dependsOn` reduce needs every unit at once. Implementation uses a pipeline because an independent unit can be under review while another is still being written; a barrier there would waste the fast units' wall clock.

**Spec drift is reported, never silent.** An agent that must deviate does the code-correct thing and records a `specDeltas` entry with the spec line and the code evidence. Reviewers are told declared deltas are expected and that *undeclared* divergence is a review failure. That is how the spec and the code converge instead of quietly forking. (The specs already produced one live example: the W0 migration number was wrong because `0020` shipped the same morning.)

**Nothing writes to GitHub.** No workflow pushes or opens a PR. The caller owns that choke point, downstream of the human approval. The three-script split exists because a Workflow script cannot call `AskUserQuestion`, so each human gate forces a boundary.

## Validation

- All three scripts parse (`node --check` after the standard async wrapper — top-level `return` is Workflow idiom, as in the existing `issue-*.workflow.mjs`).
- Manifest DAG verified: 11 units, 0 dangling dependencies, no cycles. Topological order runs `w0-schema → w0-store → … → w0-wire`.
- All files under 500 lines.

## Not in scope

No product code for the dev platform itself. W0 implementation starts from this manifest once a human approves it.

Refs #470


Weegy:
Heads-up: merged current `main` into this branch and renumbered the dev-platform migration **0021 → 0022** (commits 30f5915, 73f48b8, 2b130a5). PR #477 merged first and took `0021_plugin_verdict.sql`, so this PR's migration had to move — per the collision note in both PRs. All code/doc references (`types.ts`, `devJobStore.ts`, `devJobArtifactStore.ts`, `devRunnerApi.ts`, `w0-manifest.json`) updated to 0022. Later dev-platform waves that reserved 0022–0024 in the epic #470 specs shift accordingly (now 0023–0025). Uncommitted work-in-progress in the local worktree (launch-ACL changes in `devPlatform.ts` + tests) was left untouched.

PR #538, opened 2026-07-29
docs(470): plugins use Tailwind, so they ship no CSS (G7 reduced)

Weegy:
Follow-up to #536. Documentation + a measured reference artifact — no runtime code.

## The idea

Marcel's, and it's better than the workaround it replaces: web-ui is a Tailwind v4 project, so **require Tailwind in plugin markup and a plugin never needs to ship CSS at all** — it links a stylesheet core serves.

This matters for the Dev Platform extraction (`specs/470-dev-platform-plugin/plan.md`), where G7 — "a distributed plugin cannot ship a React UI" — was rated a hard blocker partly because the ZIP extension allowlist has no `.css`. That turns out to be the wrong thing to fix.

## The catch, which is the whole design

**Tailwind emits only classes it has seen.** v4 detects them by scanning source files at build time, and a plugin installed at runtime from another repository is never scanned. So "plugins use Tailwind" only works if core **pre-generates a documented, finite vocabulary**.

v4 supports exactly that: `@source inline(...)` — the replacement for v3's `safelist`, brace-expandable — combined with `@import "tailwindcss" source(none)` to disable scanning, producing a stylesheet containing precisely that set and nothing else.

## Measured, not estimated

Built with the repo's own `tailwindcss@4.3.3` + `@tailwindcss/postcss`. Probe kept at `specs/470-dev-platform-plugin/plugin-tailwind-subset.probe.css` (reference artifact — not shipped, not built by web-ui):

| | |
|---|---|
| Vocabulary | layout, flex/grid, spacing 0–12, typography, borders, shadows, `sm:`/`md:`/`lg:`, `hover:`/`focus:`/`disabled:` |
| Colours | **Lume tokens only** — `.bg-accent`, `.text-fg-muted`, `.border-border`, `.text-danger` verified present in the output |
| Size | 43,199 B raw → **7,704 B gzip** → 5.7 KB brotli |

## Why this is worth more than unblocking one extraction

1. **Plugins inherit the design system by construction.** They get *our* colour names wired to the runtime CSS variables, so they follow the active palette and light/dark automatically. A plugin cannot hardcode a hex and drift.
2. **It retires a known drift hazard.** `middleware/src/admin-ui/harness-admin-css.ts` is 345 hand-maintained lines whose own header says *"mirror `web-ui/app/_lib/theme.css`; keep the two roughly in sync when the design system changes."* Generating both from the same tokens removes the sync obligation instead of restating it.
3. **It is enforceable cheaply** — reject `[` in class attributes at package ingest.

## The constraint this puts in the plugin contract

**No arbitrary values** (`w-[137px]`, `bg-[#abc]`). That space is unbounded and cannot be pre-generated, so such a class renders **unstyled with no diagnostic** — the worst failure mode. Documentation alone won't hold it; it needs the ingest check.

Implementation note: the `@theme inline` token bridge currently lives inside `globals.css:16-48` and must be **extracted to its own file** that both it and the plugin stylesheet import — otherwise the two drift, which is the exact failure this is meant to end.

## Effect on the plan

G7 drops from "hard blocker" to the JS-bundle question alone: `.js` and `.map` are already allowlisted, so a compiled SPA can ship today; what is still missing is a static-asset serving path from the plugin's router. Phase P3b is re-scoped accordingly.

## Test plan

- [x] Probe compiles with the repo's own Tailwind toolchain; output inspected for the Lume colour utilities and the `sm:`/`md:`/`hover:` variants
- [x] Sizes measured (raw / gzip / brotli)
- [x] No change to any shipped file — the probe lives under `specs/`, is not imported by `web-ui`, and is not part of any build
- [ ] Implementation itself (extract the `@theme` bridge, build + serve the sheet, ingest check) is P3b, not this PR

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/538"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787898674&installation_model_id=428086&pr_number=538&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F538&signature=e02f3fdc6c58dad3d475cb59bd211732a2eae5e89811f08aea42ace28f35e75b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,31 @@ permissions:
 
 jobs:
   # ------------------------------------------------------------------
+  # Core-decoupling ratchet (epic #470). The Dev Platform is being
+  # extracted into its own repository; this fails if core re-acquires
+  # references to it. The count may fall freely and never rise without
+  # a hand-edited baseline, so the extraction cannot silently regress
+  # while it is in flight — and "finished" is machine-checked (count 0)
+  # rather than asserted.
+  # ------------------------------------------------------------------
+  decoupling:
+    name: core decoupling ratchet (#470)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version: '22'
+
+      - name: Ensure ripgrep
+        run: command -v rg >/dev/null || (sudo apt-get update && sudo apt-get install -y ripgrep)
+
+      - name: Check core is not re-coupling to the Dev Platform
+        run: node scripts/check-core-decoupling.mjs
+
+  # ------------------------------------------------------------------
   # Middleware: lint + typecheck + node:test against tsx
   # ------------------------------------------------------------------
   middleware:
--- a/scripts/check-core-decoupling.mjs
+++ b/scripts/check-core-decoupling.mjs
@@ -1,0 +1,229 @@
+#!/usr/bin/env node
+/**
+ * Core-decoupling ratchet for the Dev Platform extraction (epic #470).
+ *
+ * WHY THIS EXISTS
+ * ---------------
+ * `specs/470-dev-platform-plugin/core-decoupling-checklist.md` is a snapshot: it
+ * tells you what was coupled the day it was written, and it goes stale the
+ * moment anyone touches the tree. A checklist cannot tell you whether the
+ * extraction is *finished*, and it cannot stop core from quietly re-acquiring a
+ * dependency while the work is in flight.
+ *
+ * This does both, by counting references and refusing to let the count rise.
+ * Extraction is done when the count is zero — that is the definition, and it is
+ * machine-checked rather than asserted.
+ *
+ *   node scripts/check-core-decoupling.mjs            # verify against baseline
+ *   node scripts/check-core-decoupling.mjs --report   # per-zone breakdown
+ *   node scripts/check-core-decoupling.mjs --update   # lower the baseline
+ *
+ * `--update` only ever lowers it. Raising the baseline requires editing the
+ * committed value by hand, which is exactly the kind of change that should
+ * show up in a diff and be argued for in review.
+ */
+
+import { readFileSync, writeFileSync, existsSync } from 'node:fs';
+import { execFileSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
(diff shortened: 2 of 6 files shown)


Weegy:
Scope grew — this PR now carries two things, both independently reviewable:

1. **Tailwind for plugins** (`012daf3`) — docs + the measured probe. Plugins use Tailwind, so they ship no CSS; core pre-generates a finite vocabulary via `@source inline(...)`. 7.7 KB gzip carrying the Lume tokens.

2. **Decoupling ratchet + acceptance matrix** (`b145620`) — answers a question the plan could not: how do we know every function got extracted and that it installs?

Added because `core-decoupling-checklist.md` enumerates *files* — you can move all ~200 and still silently lose a feature. The additions:

- `scripts/check-core-decoupling.mjs`, wired in as the CI job **core decoupling ratchet (#470)**. Counts Dev Platform references across 12 zones and fails if the count rises. Baseline **3,171**; `--update` only lowers it. Makes "finished" machine-checked (count 0) rather than asserted, and stops core re-acquiring a dependency mid-extraction. Verified in both directions.
- `specs/470-dev-platform-plugin/acceptance.md` — the functional contract: 34 endpoints, 3 chat tools, `ctx.devJobs`, 4 background loops, 4 UI screens, the chat card, the CLI and the conductor step kind, each with a probe, plus install/uninstall/upgrade acceptance.

They're stacked because both edit `plan.md`. Happy to split into two PRs if you'd rather review the CI change on its own.

Weegy:
Superseded by #539 — the epic now lives in a single PR so planning and implementation share one home, as requested.

Both commits here are carried over unchanged (`012daf3` Tailwind, `b145620` ratchet + acceptance). Nothing is lost; closing this to avoid two open PRs against the same epic.

PR #618, opened 2026-08-05
fix(chat): route background stream writes by session id, not active t…

### LINKED ISSUE #617: Background chat turns lose their content when the user switches tabs mid-stream
## Summary

A background chat turn loses its content if the user switches tabs while it is streaming. The stream keeps running and its tab reports "response ready", but the answer never lands in the session.

**Pre-existing on `main`** — not introduced by #409. Filing it because #409's per-tab markers make the failure *legible* (a tab now advertises a finished answer that isn't there), and because `chatStreamEvents.ts` already documents the limitation in a comment rather than tracking it as a bug.

## Mechanism

`StreamRunner` folds every event through `applyStreamEvent`, which writes via `sessions.mutateActive`:

```ts
// web-ui/app/_lib/chatStreamEvents.ts:168
sessions.mutateActive((session) => {
  if (session.id !== sessionId) return session;   // ← no-op for a background session
  return foldEvent(session, pendingMessageId, event);
});
```

The file says so itself (`chatStreamEvents.ts:156-160`):

> Note: mutateActive only writes to the *currently active* session. […] If we ever want to run background turns for a non-active session we'll need a per-session `mutateById` helper.

`finalizePending` (`StreamRunner.tsx:245-250`) has the same shape, and `persistActive()` (`:241`) persists whichever session is active at that moment — not the one the turn belongs to.

## Repro

1. Open two chat tabs, A and B.
2. Send a message in A.
3. Switch to B before A finishes.
4. Return to A.

**Expected:** A holds the complete answer.
**Actual:** A's message is truncated at the point of the switch and can remain stuck in `streaming`; the authoritative `done` answer is discarded. A's tab still shows the "response ready" marker, and pre-#409 the toast said "Answer ready" for the same turn.

## Suggested fix

Expose `mutateById(sessionId, mutator)` and `persistById(sessionId)` on the chat-sessions store and route every stream event and finalization through them, rather than through the active-session helpers. That is the `mutateById` helper the existing comment anticipates.

## Notes

- Found by a cross-vendor (GPT-5) review pass on #409.
- Worth a regression test that drives a turn to `done` for a non-active session and asserts the message content landed.

sneumannb5:
## What
  Closes #617. Background chat turns now fold every event — deltas and the final `done` answer — into the session that owns them, instead of only ever writing to the active tab.

  ## Why
  Stream writes went through `mutateActive`/`persistActive`, which by design only touch the currently active session. Switching tabs mid-stream discarded the turn's deltas and its authoritative `done` answer, leaving the message truncated and stuck `streaming` while its tab still advertised "response ready". Added `mutateById`/`persistById` (the per-session primitive the old code comment anticipated) and routed all stream writes + finalization + persistence through them; `mutateActive`/`persistActive` now delegate to them.

  ## Test plan
  - [x] `npx tsc --noEmit` — clean (only pre-existing generated `.next/dev-button-check` noise)
  - [x] `npm run lint` — 0 errors (38 pre-existing warnings, unrelated files)
  - [x] `npm run test` — 69 files / 578 tests pass
  - [x] `npm run i18n:check` — OK, 3327 keys, en+de parity
  - [x] Added regression test: drives delta+`done` into a **non-active** session, asserts the full answer lands, `streaming:false`, active tab untouched, `persistById(bgId)` called and `persistActive` never called (assertions mutation-guarded)

  ## Risk / blast radius
  `mutateActive`/`persistActive` refactored to delegate to the new by-id helpers — active-tab behavior is unchanged, and the 3 `chat/page.tsx` callers are untouched. No schema, public API, CI, or env-var changes. UI-only; no new user-facing strings.

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/618"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788528080&installation_model_id=428086&pr_number=618&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F618&signature=972d90cfe4933a944a64faacb78741b9a6ec4e0eaee9452e3388b2bdf3edef09"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/web-ui/app/_components/StreamRunner.tsx
+++ b/web-ui/app/_components/StreamRunner.tsx
@@ -238,7 +238,10 @@ export async function runOneTurn(
     }
   } finally {
     finalizePending(depsRef.current.sessions, sessionId, pendingMessageId);
-    void depsRef.current.sessions.persistActive();
+    // Persist the turn's OWN session, not whichever tab is active now — the
+    // user may have switched away while this background turn was streaming
+    // (#617).
+    void depsRef.current.sessions.persistById(sessionId);
   }
 }
 
@@ -247,18 +250,15 @@ function finalizePending(
   sessionId: string,
   pendingMessageId: string,
 ): void {
-  sessions.mutateActive((session) => {
-    if (session.id !== sessionId) return session;
-    return {
-      ...session,
-      messages: session.messages.map((m) =>
-        m.id === pendingMessageId && m.streaming
-          ? { ...m, streaming: false, finishedAt: Date.now() }
-          : m,
-      ),
-      updatedAt: Date.now(),
-    };
-  });
+  sessions.mutateById(sessionId, (session) => ({
+    ...session,
+    messages: session.messages.map((m) =>
+      m.id === pendingMessageId && m.streaming
+        ? { ...m, streaming: false, finishedAt: Date.now() }
+        : m,
+    ),
+    updatedAt: Date.now(),
+  }));
 }
 
 interface AccumulatorState {
--- a/web-ui/app/_components/__tests__/StreamRunner.test.tsx
+++ b/web-ui/app/_components/__tests__/StreamRunner.test.tsx
@@ -1,7 +1,11 @@
 import { act, render } from '@testing-library/react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
-import type { UseChatSessionsResult } from '../../_lib/chatSessions';
+import type {
+  ChatSession,
+  Message,
+  UseChatSessionsResult,
+} from '../../_lib/chatSessions';
 import {
   StreamStoreProvider,
   useStreamStore,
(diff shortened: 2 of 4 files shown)


Weegy:
Thanks for this — closing as superseded, not because it was wrong.

#617 was fixed on `main` by #620 (`9b6f656f`, merged 2026-08-06), which landed the same core idea: `mutateById` / `persistById` as the addressing primitive, with every stream write, the finalization, and the persistence routed through them.

One substantive difference is worth flagging, because it is the half of #617 this PR leaves open:

`persistById` here reads its snapshot from `sessionsRef.current`, which is synced by a `useEffect`. In `runOneTurn`'s `finally` the two calls are synchronous and adjacent:

```ts
finalizePending(...);                 // schedules setSessions
void sessions.persistById(sessionId); // reads sessionsRef.current — pre-commit
```

The ref has not been re-synced yet, so the PUT carries state that predates the `done` fold: the truncated content, `streaming: true`. For a foreground turn that self-heals, because the next user turn in the same session re-persists. A background turn has no next turn — that stale PUT *is* the final persisted state, so the bug survives a reload. #620 replaced the ref read with a queue drained from an effect against post-commit `sessions`, and additionally drops the PUT when the session was deleted mid-flight (so a late background turn cannot resurrect a discarded session).

#620 also removed `mutateActive` / `persistActive` outright and migrated the three `app/chat/page.tsx` call sites, rather than keeping them as delegating wrappers — one addressing mode, nothing to drift.

Current coverage on `main`: 3 tests in `StreamRunner.test.tsx` (`#617` block), plus `chatSessions.persistById.test.ts` and `chatStreamEvents.test.ts`. The post-commit assertion in `persistById.test.ts` is the one that pins the behavior described above.

All four files this PR touches now conflict with `main`, and resolving them toward `main` leaves an empty diff — there is nothing left to land. Closing.

Issue #333, opened 2026-06-18
Identity & Role Projection — omadia as a team-player in the existing IT landscape (Principal = user|role everywhere)


iret77:
## Summary & vision

Make **omadia an aggregator/projector of identity and roles, not an owner of them.** omadia
is not an ERP and must not become one — but ERPs, HR systems, and IdPs (which already are
the org's systems of record for *who exists* and *who is responsible for what*) dock onto
omadia. omadia should **project a live user/role view from those sources**, joined on a
primary key, rather than maintain a second/third copy that has to be kept in sync. This is
the concrete expression of omadia's differentiator: it is itself a **team-player in the
company's existing IT landscape**, not another island.

This is **already true for authentication** (see "What already exists"). This issue extends
the same "project, don't own" pattern from *access identity* to *org roles + employee
attributes*, and establishes a **platform-wide paradigm** for how any module or agent
addresses a person.

> Foundation for: **omadia Conductor** (#321, role-addressed steps & reporting) and the
> **omadia Facilitator** (#330, report/escalation to a user or role). Both should **consume**
> this layer instead of owning role tables.

## The paradigm (headline)

**Principal = `user:<id>` | `role:<key>` is the DEFAULT addressee type everywhere in omadia.**

Any omadia module or agent that targets a person as the destination of *something* — an
escalation target, a report/notification recipient, an approval/decision step, an assignment,
a hand-off — MUST by default accept **either a specific user or a role**.

- Restricting a target to **user-only** (no role) is the **exception**, permitted ONLY when
  it is **technically or legally necessary** to deal with one identified natural person
  (e.g. a legally binding personal sign-off, a per-person secret, a regulatory 4-eyes record
  tied to named individuals).
- Such a user-only restriction MUST be **explicitly justified and documented at the point of
  definition** — mirroring the existing `multi_instance_justification` precedent (a manifest
  that sets `multi_instance: false` must say why; see `specs/001-multi-orchestrator-runtime`).
  The default is role-capable; opting out costs a justification.
- Rationale: people change roles, go on leave, leave the company. Hard-wiring a workflow to a
  named person is brittle and a maintenance burden; addressing a **role** (late-bound to the
  current holder) is the resilient default. The `user`-only case should be rare and deliberate.

Concretely, this paradigm governs (non-exhaustive): Conductor human-step addressees,
Conductor escalation/fallback targets, report and interim-status recipients, notifications,
assignments, and any future "send to / ask / require sign-off from <someone>" surface.

[...]

Issue #665, opened 2026-08-12
Knowledge-graph plugin close() ends the kernel-wide pg pool — any reactivate() bricks the instance until restart, and /health cannot see it


Weegy:
## Summary

The kernel resolves the Postgres `Pool` **once** and shares that single reference with ~40 subsystems. The `@omadia/knowledge-graph-neon` plugin's `close()` calls `graphPool.end()` on it. Any code path that deactivates or reactivates that plugin therefore kills the pool **process-wide** while the process stays up and healthy-looking. Every DB-backed feature then fails permanently until someone restarts the machine.

This exact failure mode is already documented in the codebase — `routes/adminEmbeddingProvider.ts` and `packages/harness-knowledge-graph-neon/src/gateReevaluation.ts` describe it at length and explain why #440 / #537 stopped the *embedding-provider switch* from reactivating the plugin. That fix was scoped to one caller. The underlying hazard is still reachable from every other caller, and there is no guard.

Observed in production-shaped conditions on a freshly provisioned instance, ~90 minutes after a clean deploy.

## Impact

Once triggered, the instance is bricked until restarted. Log excerpt:

```
[chat-sessions] list failure: Error: Cannot use a pool after calling end on the pool
    at async ChatSessionStore.list (packages/harness-orchestrator/dist/chatSessionStore.js:29:15)
[operator-agents] Error: Cannot use a pool after calling end on the pool
    at async ConfigStore.listAgents (packages/harness-orchestrator/dist/registry/configStore.js:121:26)
[middleware] schedule worker: list failed { error: 'Cannot use a pool after calling end on the pool' }
[@omadia/orchestrator] reloadBus: reload FAILED {"source":"reconcile","error":"Cannot use a pool after calling end on the pool"}
[session-log] markdown write failed: Cannot use a pool after calling end on the pool
    at async PostgresMemoryStore.writeFile (packages/harness-memory-postgres/dist/postgresMemoryStore.js:78:13)
```

[...]

Issue #565, opened 2026-07-31
CI never runs any `.pg.test.ts` — the middleware job has no postgres service


Weegy:
Only the `schema (migrations on pgvector)` job declares a postgres service
(`.github/workflows/ci.yml:146-148`). The `middleware` test job does not, so **every
Postgres-backed suite silently skips in CI** — the migration backfill tests, the embedding-gate
suites, `devJobTaskStore.pg.test.ts`, and the rest.

Silent skipping is the dangerous part: the job reports green and the suites look covered.

Surfaced during PR #550, where a restored migration assertion could only be verified locally, on a
privately-run `pgvector/pg16`.

Evidence that switching it on is probably cheap: a full local run **with** postgres enabled on node
22.22.3 gave 5982 tests / 5972 pass / 6 fail / 4 skip in 169 s, and none of the six failures were
Postgres-related (`devplatform/goldenFixture.e2e`, `builder preview routes`, runner-router
credential tests — all local-environment `ENOENT`/worker-provisioning artefacts).

Deliberately **not** fixed inside #550: it changes CI infrastructure for the whole repo, would be
invisible inside a 116-file MCP change, and if it does red-line on pre-existing failures it blocks a
PR that is otherwise ready.

_Filed as a follow-up from PR #550 (MCP 2026-07-28 cluster, wave 4–6)._

PR #691, opened 2026-08-13
fix(#568): bridge channel turns to per_user MCP tokens via the IdP subject

### LINKED ISSUE #568: Channel turns cannot reach a `per_user` MCP token — identity-namespace mismatch
PR #550 (W4-1) added the missing `mcpUserKey` producer. The HTTP chat path is end-to-end functional.
The channel path resolves an identity and attributes it correctly in the audit trail, but **keys on
the wrong namespace**: channel turns resolve the canonical omadia uuid, while `/authorize` stores
tokens under the session-shaped identity (`agentBuilder.ts:1455` →
`resolveMcpUserKey(server, sessionIdentity(req))`, i.e. `session.sub || session.email`).
`session.sub` is `providerUserId` (`routes/auth.ts:619`) — the AAD `claims.oid` for entra
(`EntraProvider.ts:154`), the users-row `providerUserId` for local
(`LocalPasswordProvider.ts:96`).

So an affected user fails closed with a clear message rather than silently borrowing the operator's
identity — the safe half — but still cannot reach their `per_user` server from a channel.

**Why it could not be closed in #550.** The chain channel-native id → session-shaped key is not
reachable through the public contract:
1. The `KnowledgeGraph` interface exposes exactly one identity method,
   `resolveOrCreateChannelIdentity` (`plugin-api/src/knowledgeGraph.ts:196`). Nothing walks a cluster.
2. `ResolveOrCreateChannelIdentityResult` (`:1206-1217`) returns `channelIdentityNodeId`,
   `userNodeId`, `omadiaUserId`, `isNewIdentity`, `isNewCluster` — no email, no `aadObjectId`, no
   sibling identities. There is no way to reach the cluster's `web` ChannelIdentity, whose
   `channelUserId` is the local `users.id` (`index.ts:3270-3272`) and the only bridge to
   `provider_user_id`.
3. The neon implementation exposes no un-interfaced helper to shortcut the contract.
4. The tempting shortcut — treat the Teams `channelUserId` as the AAD oid — is unsound.
   `ChatParticipant` (`chatParticipants.ts:9-19`) explicitly separates `channelUserId`
   ("used verbatim as `mentioned.id`") from `aadObjectId`. `chat.ts:81`'s comment claims the
   opposite, and the Teams plugin that would settle it lives in the private
   `omadia-byte5-plugins` repo. Keying MCP tokens on an unverifiable premise is the same class of
   mistake as the `x-user-id` header fall-through W4-1 avoided.

Closing it needs a new method on the `KnowledgeGraph` public contract, implementations in
`knowledge-graph-inmemory` and `knowledge-graph-neon`, and new boot wiring in `index.ts` to reach
`UserStore`.

**Scope is narrower than it first looks.** A `per_user` token can only exist for a human who
completed `/authorize` in the admin UI, and that flow requires a session. A Teams-only user has no
users row, no session, and therefore no token — failing closed is complete and correct for them.
The affected population is people who are *both* an admin-UI user holding a `per_user` token *and*
a channel user.

**A decision this issue must make and #550 deliberately did not:** what should happen to
Telegram/Slack users who have no admin-UI account at all.

Weegy:
## What

A channel turn (Teams/Telegram/Slack) keyed its MCP identity on the KG-canonical omadia uuid. `/mcp-servers/:id/authorize` stores a `per_user` OAuth token under the session's `sub` — the auth provider's own subject. Two namespaces that never met: a per-user token an operator granted in the Admin UI was never found from a channel, and every such call failed closed.

The bridge really was absent — `resolveOrCreateChannelIdentity` returned ids only, so nothing downstream could recover the subject.

## How

| Layer | Change |
|---|---|
| `plugin-api` | `ChannelIdentityIngest.authSubject` records which IdP subject authenticated; the resolve result gains `clusterAuthSubject`. `authSubjectProps` lives here so both backends persist it under identical property names |
| `neon` + `in-memory` | Persist the subject, **backfill on the fast path** (every login re-enters there, so pre-existing identities are not stranded), and **merge rather than replace** so a channel-side re-resolve cannot erase a subject a login established |
| `resolveTurnOwnerIdentity` | Returns `{ omadiaUserId, authSubjectKey }`, carried by the single round-trip the turn already pays for — no second identity lookup |
| `orchestrator` | Both `mcpUserKey` producers prefer the subject; canonical uuid stays the fallback |
| `src/index.ts` | The login flow records `authSubject`, behind a completed authentication |

## Why an explicit fact, not an inference

For the local provider `providerUserId` *happens* to be the lowercased email today, so the subject could have been inferred from `email`. That is exactly the kind of convention that rots silently in a credential path, so it is stored as its own field. The Neon prop schema declares the keys rather than leaning on `.passthrough()`, for the same reason `resultType` is declared in the MRTR client schema: passthrough residue reads as incidental, and a future tightening would break the bridge silently instead of loudly.

## Security

Not a trust change. Both values are KG-attested and neither is client-controlled — this is a namespace correction, not a downgrade. Absence of a subject still means "no token to inherit", never a substitute key, so W0-1's confused-deputy fix is preserved and a channel-only user's behaviour is byte-identical to before.

**Known limit, deliberately not papered over:** a cluster that merged a local-password login and an Entra login holds two subjects. The pick is deterministic (most recently seen); a token authorized under the *other* one is not found and the call fails closed. Widening that means teaching the token lookup to try an ordered key set — a change to the credential read path, out of scope here.

## Two tests rewritten, with reason

`test/orchestrator/mcpUserKeyProducer.test.ts` asserted `perUser.userKey === CANONICAL_UUID` as the desired outcome. That pinned the bug: they were green only because they checked that *some* key was produced, never that it was a key a token could exist under. Same shape as the #550 security test that pinned the vulnerable variant. The canonical uuid is still asserted — as the fallback it is, not as the goal.

## Verification

- **Mutation check (producer)** — reverting to `resolvedOmadiaUserId` and rebuilding `dist/`: 2 of 14 red (exactly the bridge tests; the fallback test correctly survives). Restored, rebuilt, 14/14 green.
- **Mutation check (graph)** — forcing `clusterAuthSubject` to `{}` and rebuilding `dist/`: 4 of 28 red (the two negative-case tests correctly survive). Restored, rebuilt, 28/28 green.
- `npm run build` clean; `npm run typecheck` clean; `npm run lint` clean.
- Test-typecheck ratchet: **406, baseline 406** — not raised (one real regression this change introduced in `queryDatasetTool.test.ts` was fixed, not baselined).
- Decoupling ratchet: held at **3294**.
- 315 tests green across the touched surfaces (slice-1b, MCP user-key producer, run-trace ingest, KG walk/merge/topic/inconsistency, dev-graph router, MCP OAuth, session logger, auth suite).

## Neon executed, not just shape-checked

A second commit adds `test/channelIdentityAuthSubject.pg.test.ts`: 6 tests driving the **real Neon implementation** against pgvector/pg17, including a cross-tenant leak case. The in-memory suite could not have caught the fast-path `jsonb` merge, the `IS_IDENTITY_OF` traversal direction, or the in-transaction read-back — all SQL-shaped.

Mutation check: dropping `|| $3::jsonb` from the fast-path UPDATE and rebuilding `dist/` turns the backfill test red (1 of 6); the other five correctly survive. Restored, rebuilt, 6/6 green.

The var chain includes `GRAPH_PG_TEST_URL` / `MEMORY_PG_TEST_URL`, both set by `ci.yml`, so the suite **runs in CI** rather than skipping into a permanently-green no-op.

_Note on the first CI run: `middleware (lint + typecheck + test)` failed on `a RESUMED task survives the reaper` (TaskStore pg, #561) — a latent timing race unrelated to this change (60 ms stale window against a `sleep(120)` with a DB round-trip in between). A re-run of the unchanged commit was fully green._

## Scope boundary

Touches the same identity seam as #333 (Identity & Role Projection). This does **not** anticipate it: no roles, no tenancy, no projection layer — it only makes one credential table agree with an id that already existed on both sides. #333 would build on this rather than redo it.

Closes #568


Weegy:
## Follow-up I could not push: `PG_TEST_FLOOR` has drifted

Verified from this PR's CI run: the new pg suite **does** execute — the `Test — Postgres suites` step reports `ran=261 skipped=0 (floor=240)`. (The `no test Postgres configured` line in the log is the *other* step, plain `npm test`, which by design carries no DSN — every one of the 25 pg suites skips there.)

While confirming that, the floor turned out to be stale. Before this PR the pg suites registered **255** tests against a floor of **240**, so any single suite could silently self-skip inside 15 tests of slack without going red — which is exactly the failure `ci.yml` says this guard exists to catch:

> only a floor on the test COUNT catches it … If the count drops (broken DSN, a future suite reading an env var we don't set here, a cold-connection flake), a suite silently vanished — go red.

This PR takes it to 261. Suggested one-line change, per the file's own "Raise `PG_TEST_FLOOR` when you add Postgres tests":

```yaml
-          PG_TEST_FLOOR: '240'
+          PG_TEST_FLOOR: '261'
```

and the comment above it updated from "register 240 tests today (+2 from `kgPluginPoolLifetime.pg.test.ts`, issue #665)" to "register 261 tests today (+6 from `channelIdentityAuthSubject.pg.test.ts`, issue #568)".

Worth stating explicitly in that comment, because the two ratchets in this repo pull in opposite directions and the wrong reflex is easy: this floor is a **minimum on desirable coverage**, so raising it as suites land is correct. The core-decoupling ratchet counts **debt**, where the rule is reword-don't-raise.

**Why it is not in this PR:** the change touches `.github/workflows/`, and neither push route is available to me here — the SSH agent has no identities, and the `gh` token carries `admin:public_key, gist, read:org, repo` but **not `workflow`**. I left the branch at exactly the two commits CI went green on rather than carry an unpushable third. Happy to apply it once a key is loaded, or it is a safe one-liner to land directly.

Weegy:
**Correction to the comment above: the floor fix is now in this PR** (`b8c0395e`).

The push failure was transient — an empty `ssh-agent`, not the missing `workflow` scope I attributed it to. SSH authenticates fine now, so the change landed here rather than being handed off.

The number also moved while this sat: the floor is pinned to **269**, not the 261 suggested earlier. This branch's CI reports `ran=269 skipped=0 (floor=240)` now that #692 (#432, rolling self-update) has merged and brought `updateAuditStore.pg.test.ts` with it — **+8** on top of this PR's **+6**.

So the drift was worse than first reported: **29 tests of slack**, in which an entire pg suite could self-skip without going red. That is exactly the #565 failure the guard exists to catch, and two PRs in a row widened it by adding suites without moving the floor.

Taken from the step's own `ran=` output rather than estimated, and the comment now records where to read it. It also spells out that this floor is the *inverse* of the core-decoupling ratchet — a minimum on desirable coverage, where raising is correct, versus a debt counter where the rule is reword-don't-raise. Applying the wrong reflex here would quietly disarm the guard.

_Also in this PR since the last CI green: `origin/main` merged in (#691 had gone `behind` once #692 landed). #692 touched `middleware/src/index.ts`, the same file as this fix, so the merge was re-verified rather than trusted — full rebuild, the #568 suites (47 tests) and the Neon pg suite (6) re-run green._

Issue #343, opened 2026-06-20
Lume spec updated in omadia-ui (icons + state-as-glyph) — review core implications


iret77:
**Notice — the Lume specs in `byte5ai/omadia-ui` were updated; please review core implications.**

Merged to `omadia-ui:main`:

- **Icons** — canvas-protocol §12 (v1.1 additive): `icon` trait + `IconRef` (`app:`/`lib:`/`gen:`), `iconResolver@1`, deferred `iconGenerator@1`; `visual-spec` §2.12 rewritten → v0.6. (byte5ai/omadia-ui#41)
- **State-as-glyph** — "show state, don't narrate it" + a `chart` spark variant. (byte5ai/omadia-ui#42)

**To review in core (Tier 2/3 + SDK + Skill):** SDK types (`icon` trait / `IconRef` / chart `variant`), orchestrator composing icons, UI-Skill authoring rules ("show state, don't narrate" + icon placement + the curated `app:` subset), deferred `iconGenerator@1`.

Specs: `byte5ai/omadia-ui` → `docs/iconography.md`, `docs/data-glyphs.md`, `docs/protocol/1.0.md` §12.

These are proposals — a review pointer, not a committed implementation spec.

Weegy:
## Triage
**Priority:** P2 · **Category:** architecture · **Effort:** M · **Status:** Needs decision — adopt-in-core call pending; upstream specs are still RFC

Verified against `main@1e5b24d`: nothing from this has landed in core. `IconRef`, `iconResolver`, `iconGenerator`, an `icon` trait, and a spark/compact `chart` variant have zero occurrences in `middleware/` and `web-ui/`; the last substantive canvas commits (`8c10ccb` Lumens 1.1 #315, `9920b53` #324) predate this issue. The referenced omadia-ui docs (`docs/iconography.md`, `docs/data-glyphs.md`) explicitly say "proposal / RFC … not yet merged into either canonical spec", so the core-side question is whether to adopt the additive protocol changes now or wait for spec canonization.

## Implementation notes
**Affected code:**
- `middleware/packages/canvas-core/schema/canvas-tree.schema.json` — `commonTraits` ($defs) is where the `icon`/`iconState` trait goes; every primitive merges it via `allOf` + `unevaluatedProperties:false`, so today an `icon` prop is **rejected** by the whitelist. `p_chart` currently allows only `chartType: bar|line|pie` — the spark variant is an additive prop here.
[...]

PR #725, opened 2026-08-17
feat(#343): adopt Lume icon + state-as-glyph specs (additive, gate-re…

sneumannb5:
## What
  Adopt the Lume icon + state-as-glyph specs (#343) into core as an **additive, gate-ready** change. Adds the `icon`/`iconState` trait and chart `variant:"spark"` to the canvas-tree whitelist, plus the "show state, don't narrate it" authoring rule. ADR-0010 records the decision.

  ## Why
  The upstream Lume specs (omadia-ui#41/#42) proposed an additive canvas-protocol §12 icon affordance and a spark chart variant; none had landed in core. This lands the vocabulary + validator contract so the desktop/mobile renderers have one thing to build against — without turning emission on. Nothing in core emits the new props yet; emission stays behind the `iconResolver@1` handshake gate, because the Tier-1 renderer rejects unknown props wholesale (strict `unevaluatedProperties:false` whitelist), so a half-shipped prop would brick the whole surface rather than degrade.

  ## Test plan
  - [x] `npm run gen && npx vitest run test/validator.test.ts test/fixtures.test.ts` in `canvas-core` → 41 passed
  - [x] Accept cases: `app:`/`lib:` icon + `iconState`, chart `variant:"spark"`, `_gallery.json` fixture read-back
  - [x] Reject cases: non-namespaced icon, deferred `gen:` prefix, `iconState` without `icon` (dependentRequired), unknown `iconState`/chart variant
  - [ ] Cross-repo: Tier-1 renderer `iconResolver@1` support (byte5ai/omadia-ui) — out of scope, tracked as follow-up

  ## Risk / blast radius
  - **Schema change (additive only):** new optional props on `commonTraits` + `p_chart`; no migration, no existing tree invalidated. `$id` stays `protocol/1.0/…` (additive vocabulary in the 1.0 whitelist — see ADR-0010 version story).
  - **Security:** `gen:` prefix is **rejected by the schema** (the last gate before a renderer) until the upstream §6 sanitised-SVG + consent review lands; pattern widens to include it in that same PR. `app:`/`lib:` only today.
  - **a11y:** state-as-glyph rule keeps a text/glyph label alongside `tone` — state is never signalled by colour alone.
  - No public API bump, no new env-var, no CI/release tooling touched.

  - [x] `iconGenerator@1` pinned to `generateAsset` scope `{kind:"icon"}` (not a new capability enum) — decided in ADR-0010, no churn.

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/725"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789544042&installation_model_id=428086&pr_number=725&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F725&signature=c8112ec4b453a7ad17530a37149998a40b838e4917624eba33edf59dd0bb482a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Weegy:
Prepped for merge — `main` merged in at `4ae09f74`, no conflicts, and the claims in the diff verified against the built validator rather than by reading the JSON.

### Why it needed a refresh

The 9/9 green on this branch was measured against `f1914c97`, three commits stale (#730, #748, #747 have landed since). Green against an old base is not green — so I merged current `main` in and re-ran everything.

### Verified, not assumed

The ADR says the `gen:` deferral is enforced *in the schema* ("not merely left un-emitted"). That is a security claim, so I drove it through the built validator instead of reading the pattern:

| case | result |
|---|---|
| `icon: "app:play"` | accept |
| `icon: "lib:cloud-check"` | accept |
| `icon: "gen:a-sunset"` | **reject** |
| `icon: "data:image/svg+xml;base64,…"` | **reject** |
| `icon: "app:../../etc/passwd"` | **reject** |
| `icon: "play"` (no namespace) | **reject** |
| `iconState` without `icon` | **reject** (`dependentRequired`) |
| `iconState: "active"` with `icon` | accept |
| `iconState: "hover"` | **reject** |
| `variant: "spark"` on chart | accept |
| `variant: "sparkline"` | **reject** |
| `tone: "success"` on status | accept |
| `tone: "chartreuse"` | **reject** |

13/13 as documented. The `data:` URI and traversal cases are not in the PR's own tests and are worth adding — they are the shapes an attacker reaches for once `icon` is a string that a renderer resolves.

### Test state

`main` and this branch, measured back to back in the same environment: **6830 tests, 6818 pass, 0 fail, 0 cancelled** on both. Identical — which is what we want, since the 32 added test lines are vitest in `canvas-core` (15/15 green) and run under a different runner than `middleware/test`. Test-file sets are byte-identical between the two branches (674 files, empty diff), so nothing was lost in the merge.

### Two notes for the reviewer

**1. The `iconResolver@1` gate does not exist yet.** Both schema descriptions say *"emission is gated on the renderer advertising `iconResolver@1` in the handshake"*. Repo-wide, `iconResolver` occurs only in those description strings, the generated validator that embeds them, and the ADR — **there is no handshake check in any `.ts`**. That is consistent with the PR being titled *gate-ready*, and it is safe today only because nothing emits icons: `composition.ts` gained the state-as-glyph rule and `status.tone`, but no icon instruction.

The risk is the next PR. Whoever adds icon emission must build the handshake gate in the *same* change — otherwise a Tier-1 renderer without icon support rejects the surface **wholesale**, because the whitelist is `unevaluatedProperties: false`. Worth stating in the ADR's consequences section so the sequencing is not carried only in a schema comment.

**2. Unrelated repair worth naming.** `docs/adr/README.md` gains rows for **0009 and 0010**. `0009-default-ttl-for-mcp-tool-list-caching.md` is on `main` but was never indexed in the README — this PR quietly fixes that omission from #702. Good catch; just noting it so it is not read as a merge artefact.

No blockers from my side.

Weegy:
Refreshed again — `main` merged in at `e0f6434a`. Three PRs landed since the last refresh (#752, #751, #750), which put this branch back to `mergeable_state: behind`, i.e. green against a base that no longer exists.

Re-verified on the new head:

- `canvas-core` validator suite **15/15**
- the 13-case schema probe from my earlier comment still holds — `app:`/`lib:` accepted; `gen:`, `data:image/svg+xml;…`, `app:../../etc/passwd` and namespace-less `play` rejected; `iconState` without `icon` rejected; `variant: spark` accepted, `sparkline` rejected
- full middleware suite **6860 tests, 0 fail, 0 cancelled**

One thing that is genuinely better than last time: #752 fixed the `core decoupling ratchet` job, which had been timing out inside its `apt-get` step and reporting `cancelled` while the step that does the checking was skipped. That check now completes in ~50s, so this PR's run should give a real verdict on it rather than an absent one.

Still no blockers from me. The two review notes from before stand — the `iconResolver@1` handshake gate does not exist in any `.ts` yet (safe only because nothing emits icons; whoever adds emission must build the gate in the same change), and the ADR-0009 README row this PR adds is an unrelated repair from #702 worth keeping.

Issue #482, opened 2026-07-10
Locale-aware C0 prompt-PII patterns for es/fr/nl (amounts, dates, phones) — unblock mask_user_prompt for the remaining locales


Weegy:
## Context

#361 shipped the C1 GLiNER detector tier and the committed 6-locale validation run (PR #481). The recorded run (`middleware/packages/harness-plugin-privacy-guard/src/validation/RESULTS.md`, also posted on #361) shows **de/en/it passing ALL gates** on `c0+c1`, while **es/fr/nl fail the structured-recall gate (>= 0.97)** — in every case a C0 regex locale gap, not a C1 quality problem. C1 already carries `person` at 100% recall and lifts `address` to 100% in all six locales.

## Recorded blocking gaps (from RESULTS.md)

| Locale | structured recall (c0+c1) | Gap |
|---|---|---|
| es | 95.4% | separator-less currency amounts ("899 €"), local phone formats |
| fr | 75.9% | space-grouped amounts ("2 400 €"), written-out dates |
| nl | 96.3% | dashed date format ("24-12-1987") |

## What to do

Extend the C0 regex baseline in `middleware/packages/harness-plugin-privacy-guard/src/promptMask.ts` (detector `c0-regex`) with locale-aware patterns for the recorded miss classes: amount formats (space-grouped thousands, separator-less), date formats (dashed, written-out month names for fr), and es phone formats. Keep C0 deterministic and throw-free; do not widen en/de/it patterns in ways that regress their recorded precision (>= 0.85 on negatives).

## Acceptance criteria

- [ ] Harness run (`promptDetectorEval.ts --markdown`, sets `c0` + `c0+c1`) green on ALL committed gates for es, fr, nl — without regressing de/en/it
- [ ] Native-speaker spot-check of the fr/es/it/nl fixtures (standing caveat from the validation README) resolved or explicitly re-recorded
- [ ] Results posted to the run record; only then may `mask_user_prompt` flip on for es/fr/nl per the flag policy from #361

Refs #361, PR #481.

PR #744, opened 2026-08-19
fix(#727): mask ISO-8601 dates as dates, not phone numbers

### LINKED ISSUE #727: Privacy Shield masks ISO-8601 dates as phone numbers — destroys date columns in CSV datasets at rest
## Summary

`maskPrompt` replaces an **ISO-8601 date** (`2026-07-02`) with a **phone** surrogate (`+49 30 55590000`) instead of a date surrogate. The dotted German form (`02.07.2026`) is handled correctly (`01.01.1970`), so this is specific to the `YYYY-MM-DD` shape — the shape a CSV export, an API payload, or a log line is most likely to carry.

Two surfaces are affected, with very different severity:

| Surface | Call site | Severity |
|---|---|---|
| Chat prompt masking | `harness-plugin-privacy-guard/src/service.ts:498` | Degraded model input; recoverable (the server-held pseudonym map restores on the answer side) |
| CSV dataset import (#430) | `harness-orchestrator/src/datasetImport.ts:236` | **Permanent corruption at rest** — the masked value is what gets persisted, and no map is kept |

## Repro

`middleware/iso-date-probe.ts`, run from `middleware/` with `node --import tsx ./iso-date-probe.ts`:

```ts
import {
  createBaselineDetector,
  maskPrompt,
} from './packages/harness-plugin-privacy-guard/src/promptMask.js';

for (const cell of ['2026-07-02', '02.07.2026', '+49 171 2345678']) {
  const detector = createBaselineDetector();
  const raw = await detector.detect(cell);
  const masked = await maskPrompt(cell, [detector]);
  console.log(
    `${cell.padEnd(16)} -> ${masked.maskedText.padEnd(16)}` +
      ` raw=${JSON.stringify(raw.map((s) => [s.type, cell.slice(s.start, s.end)]))}` +
      ` kept=${JSON.stringify(masked.spans.map((s) => [s.type, s.value]))}`,
  );
}
```

Actual output:

```
2026-07-02       -> +49 30 55590000  raw=[["phone","07-02"],["date","2026-07-02"]] kept=[["phone","2026-07-02"]]
02.07.2026       -> 01.01.1970       raw=[["date","02.07.2026"]] kept=[["date","02.07.2026"]]
+49 171 2345678  -> +49 30 55590000  raw=[["phone","+49 171 2345678"]] kept=[["phone","+49 171 2345678"]]
```

The date is not merely mistyped in the span metadata — it is substituted from the **phone** surrogate pool, so the output is a phone number where a date used to be.

## Root cause

The type of an exactly-tied span is currently decided by the order the patterns happen to be declared in.

1. **Two C0 patterns match the cell.** The general phone pattern (`promptMask.ts:78`)

   ```
   (?:\+\d{1,3}[\s-]?|\b0)\d{1,4}(?:[\s\-/]?\d{2,6}){1,4}
   ```

   matches the `07-02` tail through its `\b0` branch — the `-` before `07` supplies the word boundary. The ISO date pattern (`promptMask.ts:127`) matches the full `2026-07-02`.

2. **Word-boundary extension makes the two spans identical.** `WORD_CHAR` (`promptMask.ts:190`) is `/[\p{L}\p{N}_@.-]/u` — it includes `-`. So `extendToWordBoundaries` grows the phone span from `07-02` back across `-` and the digits until it covers `2026-07-02`, byte-for-byte the same range as the date span.

3. **The tie-break in `dedupSpans` falls through to array order.** The sort at `promptMask.ts:264` is

   ```
   b.confidence - a.confidence || b.end - b.start - (a.end - a.start) || a.start - b.start
   ```

   Both spans have `confidence: 1` (every C0 hit is a hard pattern match), identical length, and identical start — a **total tie on all three keys**. `Array#sort` is stable, so insertion order decides, and insertion order is the `C0_PATTERNS` iteration order, in which phone (`:78`) precedes date (`:127`).

4. **The date span is then dropped, not merged.** Phone lands in `kept` first; the date candidate overlaps it completely, `uncoveredParts` returns nothing, and it is discarded. `surrogateCandidates('phone')` supplies the replacement value.

## Impact 1 — chat prompts (recoverable, but the model reasons on garbage)

`service.ts:498` masks the turn's `request.text` with the same `maskPrompt`. Any ISO date a user types reaches the LLM as `+49 30 55590000`. Answer-side restore still works (the value is in the server-held map), so the user does not see a phone number — but the model has been asked to reason about a date it was never shown. Anything date-dependent (ordering, deadlines, "which is more recent") is being answered from a phone number.

## Impact 2 — CSV datasets (data destroyed at rest)

`datasetImport.ts:236` masks each cell and **stores the masked value**; the pseudonym map is discarded after the import. So the date is not hidden, it is gone.

Importing a CSV whose `order_date` column held ISO dates (`2026-07-02`, `2026-08-01`, …) leaves this in Postgres:

```
select jsonb_pretty(data) from dataset_rows order by row_index limit 1;

{
    "paid": true,
    "email": "lukas.becker@example.net",
    "notes": "Rahmenvertrag",
    "phone": "+49 30 55590000",
    "amount": 1299,
    "customer": "Meier GmbH",
    "order_ref": "SO-1001",
    "order_date": "+49 30 55590000"      <-- was 2026-07-02
}
```

**The stored schema and the stored data now disagree.** `inferColumnType` runs on the RAW parsed values (`datasetImport.ts:205`), so the column is correctly typed `date`; the persisted values are the masked ones. Nothing reconciles the two:

```
{ "name": "order_date", "type": "date", "sample": "+49 30 55590000" }
```

Consequence for the `query_dataset` tool: a `gt`/`lt` filter or a `min`/`max` aggregate on a column the schema advertises as `date` is evaluated against phone strings. It will not error — it will return confidently wrong answers.

Every ISO date in the column collapses to the *same* surrogate, too (each cell is masked with a fresh map — no `existingMap` is threaded at `datasetImport.ts:236`), so the rows are not even distinguishable from one another after import.

## Why "over-masking is fail-closed, so it is the lesser harm" does not cover this

The C0 comments deliberately accept false positives — e.g. the Spanish 3-3-3 phone pattern's own `HONEST CAVEAT` notes that a grouped nine-digit quantity masks as a phone, and argues a degraded prompt beats a leaked phone number. That reasoning is sound **for a prompt**, where the map makes the substitution reversible.

It does not extend to a persisted value. On the dataset path the masked text *is* the record: there is no map to restore from, the original is not stored anywhere, and the loss is silent. And the value is lost under a declared type that promises something the data no longer is.

Note also that this is not the "expected" over-masking trade: a date **is** PII by the detector's own taxonomy, and a correct date surrogate (`01.01.1970`) already exists and is used for the dotted format. Nothing here needs to be over-masked — it needs to be masked as the right *type*.

## Suggested direction (deliberately not a decision)

- **Make the precedence explicit rather than incidental.** A declared type-precedence table consulted when `dedupSpans` ties on confidence/length/start, so the outcome stops depending on where a pattern sits in `C0_PATTERNS`. Reordering the array would also "fix" this one case, and is the wrong fix — it leaves the next tie to luck.
- **Prefer the span whose match is self-contained.** The date pattern matched the whole token; the phone pattern matched a 5-char tail and only reached full coverage through word-boundary extension. "Matched natively" vs "grew into it" is available at dedup time and is a more principled key than declaration order.
- **Or constrain the general phone pattern** so an ISO date's tail cannot enter it (e.g. reject a `\b0` start that is preceded by `-` inside a longer digit run). Narrowest change, but it treats the symptom.
- **Separately, for the dataset path:** decide whether `date`-typed columns should go through the scan at all (`datasetImport.ts:229` currently sends `'string' | 'date'` through it), and whether a dataset should retain a pseudonym map so masking there is reversible like it is for a turn. Those are #430 design questions, not blockers for the masker fix.

## Acceptance criteria

- [ ] `2026-07-02` masks to a **date** surrogate, not a phone surrogate.
- [ ] `dedupSpans` resolves an exact confidence/length/start tie by a **documented** rule, and a test pins that the rule — not `C0_PATTERNS` ordering — is what decides.
- [ ] Regression coverage for the ISO shape alongside the dotted shape, so the two formats cannot drift apart again.
- [ ] The dataset path's behaviour for `date` columns is decided and documented (mask / skip / mask-with-retained-map), and the schema no longer advertises a `type` the stored values contradict.
- [ ] The C0 precision gate / `validation/RESULTS.md` is re-run, since this changes which type wins on overlapping spans.

## How this was found

Discovered while implementing #532 (the datasets admin UI). The row preview surfaced it: the page renders the stored values faithfully, and a `date` column full of phone numbers is what showed up on screen. Confirmed at rest in Postgres and then isolated to `maskPrompt` with the probe above.

No fix is included in #532 — the UI change is frontend-only and this is a middleware bug on a shared path.

sneumannb5:
## What
Closes #727 : ISO-8601 date (2026-07-02) masked as a phone surrogate. dedupSpans now breaks exact overlap ties by a documented, order-independent rule; CSV import skips date columns so stored data matches its declared type.

  ## Why
  
  Phone pattern grabbed the date's -07-02 tail, extension grew it to the full token, tying the date span — and the tie fell through to C0_PATTERNS insertion order, so phone won. On the dataset path that's permanent corruption at rest. Tiebreak is now native match length, then a fixed lexical order for determinism only.

  ## Test plan

  - [x] node --import tsx --test test/privacyPromptMask.test.ts test/datasetImport.test.ts — 60/60
  - [x] tsc --noEmit on harness-plugin-privacy-guard — clean
  - [x] C0 eval re-run — byte-for-byte unchanged (de/en 100%, es/fr/it 99.1% PASS, nl 89.0% FAIL pre-existing, precision 32/32)

 ## Risk / blast radius

  Shared path (chat masking + dataset import). Change affects only which surrogate type replaces a masked span, never whether it's masked — eval recall/precision unmoved. Documented residual: a bare-PII date column (birth_date) persists un-redacted; reversible date masking is a #430 follow-up.

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/744"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789719200&installation_model_id=428086&pr_number=744&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F744&signature=1e062cd7dcc1b4386f3fdc17a0739d34305f2ca86d05aff61badd5315e8de620"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/middleware/packages/harness-orchestrator/src/datasetImport.ts
+++ b/middleware/packages/harness-orchestrator/src/datasetImport.ts
@@ -12,15 +12,30 @@
  *
  * Scanning uses the SAME `createBaselineDetector()` (C0 regex) pass that
  * protects free-text user prompts today (`@omadia/plugin-privacy-guard`).
- * Only `string`/`date`-typed columns are scanned: a `number`/`boolean`
- * column is, by construction, a cell that parsed cleanly as a number/bool
- * for EVERY row — there is no free-text surface left for the regex to
- * match, and running it anyway risks corrupting legitimate numeric data on
- * a false-positive hit (e.g. a 7-digit id that happens to start with a
- * leading `0`, which the phone-number pattern would flag). This is a v1
+ * Only `string`-typed columns are scanned. A `number`/`boolean`/`date`
+ * column is, by construction, a cell that parsed cleanly as a
+ * number/bool/date for EVERY row — there is no free-text surface left for
+ * the regex to match, and running it anyway risks corrupting legitimate
+ * data on a false-positive hit (e.g. a 7-digit id that happens to start
+ * with a leading `0`, which the phone-number pattern would flag).
+ *
+ * `date` is skipped for the same structural reason AND a correctness one
+ * (#727): masking runs over a *persisted* value here — no pseudonym map is
+ * retained after import, so the substitution is irreversible, and a masked
+ * date would make the stored value contradict the column's declared `date`
+ * type (a `query_dataset` gt/lt/min/max over it would then compare against a
+ * surrogate string, returning confidently-wrong answers). A pure date carries
+ * no name/email/phone/address on its own, so there is nothing to redact; the
+ * real date is stored as-is and the schema stays honest. This is a v1
  * scoping call, not a bypass: every ROW still goes through the pipeline,
  * exactly as the issue requires — only cells the pipeline could not
  * possibly find PII in are skipped.
+ *
+ * Known residual (documented, not glossed): a column that is a *bare* PII
+ * date — e.g. a `birth_date` of ISO dates — is persisted un-redacted, the
+ * same class of trade as a national-ID column inferred as `number`. Reversible
+ * masking of date columns needs a retained per-dataset pseudonym map, a #430
+ * design question tracked as a follow-up, not a blocker for the masker fix.
  *
  * Cost note: this is O(rows × string-columns) baseline-detector calls,
  * each a handful of regex passes over one cell's text — CPU-bound, not
@@ -175,7 +190,7 @@ function inferColumnType(values: readonly string[]): DatasetColumnType {
 
 export interface PrivacyScanStats {
   /** Total cells (across every row) that were passed through the baseline
-   *  detector — string/date-typed columns only, see module doc. */
+   *  detector — string-typed columns only, see module doc. */
   scannedCells: number;
   /** Cells where at least one span was masked. */
   maskedCells: number;
@@ -226,8 +241,15 @@ export async function buildDatasetFromCsv(bytes: Buffer): Promise<
         outRow[header] = raw.trim() === '' ? null : /^true$/i.test(raw.trim());
         continue;
       }
-      // 'string' | 'date' — the only cells that can carry free text, so the
-      // only ones that go through the privacy scan (see module doc).
+      if (type === 'date') {
+        // Skipped like number/boolean: no free-text surface, and masking a
+        // persisted date is irreversible + contradicts the declared type
+        // (#727). Store the real date so the schema and the data agree.
+        outRow[header] = raw.trim() === '' ? null : raw;
+        continue;
+      }
+      // 'string' — the only cells that can carry free text, so the only ones
+      // that go through the privacy scan (see module doc).
       scannedCells += 1;
       if (raw.length === 0) {
         outRow[header] = raw;
--- a/middleware/packages/harness-plugin-privacy-guard/src/promptMask.ts
+++ b/middleware/packages/harness-plugin-privacy-guard/src/promptMask.ts
@@ -209,6 +209,11 @@ interface ExtendedSpan {
   readonly type: string;
   readonly detector: string;
   readonly confidence: number;
+  /** Length of the detector's OWN match, BEFORE word-boundary extension. A
+   *  span that matched its whole value natively is more self-contained than
+   *  one that only reached the same range by growing across a shared
+   *  separator — used as a tie-break key below (#727). */
+  readonly nativeLen: number;
 }
 
 /** The parts of `[candidate.start, candidate.end)` not covered by any of
@@ -240,8 +245,17 @@ function hasWordChar(text: string, start: number, end: number): boolean {
 
 /**
  * Merge detector outputs: extend to word boundaries, then resolve overlaps
- * by letting the higher-confidence span (ties → the longer span) own the
- * contested characters. A losing span is NOT discarded wholesale: the parts
+ * by letting the higher-confidence span own the contested characters. Ties
+ * are broken by a documented, order-independent rule so the outcome never
+ * depends on detector/pattern declaration order (#727): (1) higher
+ * confidence, then (2) longer extended span, then (3) larger NATIVE match —
+ * a span that matched its value directly beats one that only grew into the
+ * same range (this is what makes the ISO date `2026-07-02` beat the phone
+ * pattern that grabbed its `-07-02` tail and extended back over the `-`),
+ * then — only for two spans still identical on all three — (4) a fixed
+ * lexical order of the type name (present purely for determinism, NOT
+ * semantic priority), then (5) earliest start.
+ * A losing span is NOT discarded wholesale: the parts
  * of it no winning span covers are kept as masking spans of their own —
  * otherwise a long low-confidence C1 span (e.g. a free-form address at
  * score 0.8) that merely brushes a short confidence-1 C0 hit (the postal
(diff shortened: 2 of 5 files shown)


Weegy:
Pushed two commits to this branch (maintainer edit) while prepping it for merge, plus a branch update from `main`.

**1. `fix(#727)` — the required `middleware (lint + typecheck + test)` check was red.**

Not the test run: the #573 test/scripts typecheck ratchet. `test/privacyPromptMask.test.ts` destructured its lexical-fallback pairs from an inferred `string[][]`, so under `noUncheckedIndexedAccess` both elements were `string | undefined` and `mk()` rejected them — 4x TS2345, file not in baseline. The pairs are now a `readonly (readonly [string, string])[]`; assertions unchanged, baseline left at 406.

Worth flagging for next time: `node --test` via `tsx` strips types without checking them, so `60/60` locally and this failure are fully compatible. `npm run typecheck:test` is the one that catches it.

**2. `test(#727)` — the bug was wider than the issue described.**

The issue named only the ISO shape, but the dashed `dd-mm-yyyy` form that #482 added for nl/fr hits the identical mechanism. Straight from the raw detector:

```
2026-07-02  raw=[["phone","07-02"],["date","2026-07-02"]]
30-06-2027  raw=[["phone","06-2027"],["date","30-06-2027"]]   <- same arrangement
02.07.2026  raw=[["date","02.07.2026"]]                        <- never affected
2026-12-25  raw=[["date","2026-12-25"]]                        <- never affected
```

The trigger class is any date whose segment after a `-` starts with `0` (the phone pattern's `\b0` branch). So `30-06-2027` was persisting as `+49 30 55590000` too. I added it to the existing regression loop and confirmed it is not vacuous: narrowing that loop to only `30-06-2027` and restoring the pre-fix comparator still turns it red.

This also sharpens why RESULTS.md is right that the eval reproduces byte-for-byte — the eval scores span *coverage*, not surrogate *type*, so it is structurally blind to this entire class. Re-ran it independently and confirm your numbers: de/en 100%, es/fr/it 99.1% PASS, nl 89.0% FAIL (address-only, pre-existing since Run 2), precision 32/32 in every locale.

**Verification on the rebased head**

| Gate | Result |
|---|---|
| `npm test` (full middleware) | 6756 tests, 6744 pass, 0 fail, 0 cancelled, exit 0 |
| `npm run typecheck` / `:test` / `:golden` | exit 0 (gated on exit code, not grep) |
| `npm run lint` | exit 0 |
| C0 eval re-run | matches Run 2 |
| Mutation checks | `nativeLen`, lexical key, and the dataset date-skip each kill exactly their own test |

One note on the mutation results: removing `nativeLen` kills *only* the "native match length is the decider" test — the first ISO test stays green because for `date` vs `phone` the lexical fallback also picks `date`. That second test is doing real work; it is what keeps the comparator honest.

**Not changed, flagging instead:** the module doc says the date-column residual is "tracked as a follow-up" to #430, but there is no such issue open. The decision itself is fine and AC #4 explicitly allows `skip` — it just isn't tracked anywhere yet. Worth opening one, since a `birth_date` column now persists in the clear where it previously (wrongly, but effectively) got redacted.

Issue #687, opened 2026-08-13
i18n follow-up: the I3 literal tail and the I6 locale-formatting tail left by #679


Weegy:
Split out of #679, which covered I3 for its named worst offender plus the German-literal case, and I4/I5/I6 in full.

## Why this is a separate issue and not "the rest of #679"

**#679's I3 count does not survive measurement.** It states *25 literals across 11 files, worst offender `app/admin/kg-lifecycle/page.tsx` with 11*. A scan of `app/**/*.tsx` for JSX text nodes plus string-valued props (`title`, `placeholder`, `aria-label`, `alt`, `label`) finds **22 in `kg-lifecycle` alone** — double the figure the issue gives for that file — and the heuristic still *under*counts, because it only matches single-line JSX text. `app/graph/_components/ListView.tsx` held three German sentences that the scan missed entirely for that reason.

So I3 was never a 25-item sweep. Treating it as one would have meant closing it with most of the category still in place.

## What is left

After #679: **~217 candidates across 63 files**, from the same scan. That number is raw and deliberately noisy — a large share must NOT be translated, exactly as in #601:

| Class | Example | Verdict |
|---|---|---|
| JSON-Schema vocabulary | `string`, `number`, `integer`, `boolean`, `enum`, `array`, `object`, `minLength`, `pattern (regex)`, `format` | **never translate** — spec keywords |
| API enum values | `HOT` / `WARM` / `COLD`, `memory` / `process` / `task` | **never translate** — must match logs and SQL |
| Format-illustrating placeholders | `property_name`, `email \| uri \| date-time \| uuid`, `hostname.example.com`, `orders.created` | **never translate** — localising teaches a wrong value |
| Type-signature noise | `Promise`, `void \| Promise` | scan artefact, not a string |
| Genuine UI chrome | `Codegen`, `Preview`, `JSON syntax:`, section headings | **translate** |

The useful output of this issue is therefore the same shape #601 produced: a triaged split into *translate* vs *allowlist with a reason code*, not a bulk sweep.

### I6 tail

**18 files** still call `toLocaleString`, either bare (follows the *browser* locale, not the app's) or pinned to a tag. #679 swept `admin/usage` and `admin/kg-lifecycle`; `app/_lib/i18n-structural.test.ts` pins those two per file and can be extended one file at a time as the tail is cleared.

## Suggested approach

[...]

PR #751, opened 2026-08-19
fix(#687): triage the I3 literal tail — 21 sites, 7 files on the ratchet

Weegy:
Follow-on to #747, which made the I3 category measurable. This spends that measurement: `translate` goes from **40 → 19**, and the per-file ratchet grows from **1 → 7 files**.

## Two scanner rules, learned from the triage

Neither was guessed — both come from looking at what the first pass actually returned.

**1. Version / phase / build stamps are `diagnostic`, not prose.**

| site | literal |
|---|---|
| `store/page.tsx:229` | `omadia · v1` |
| `store/[id]/page.tsx:464` | `omadia · v1 · Slice 1.1` |
| `store/builder/page.tsx:181` | `B.0 Draft-Store` |
| `Workspace.tsx:2070` | `Phase B.5 (Workspace-UI)` |

They sit in product footers and read like text, but localising them changes an identifier an operator quotes back in a bug report. One of them is telling: `store/builder/page.tsx` already renders `{t('footer.phaseLabel')}` and then the hardcoded value right beside it — the label was translated, the value was not.

Checked in the dangerous direction: the `diagnostic` bucket contains **exactly** those four and nothing legitimate.

> Whether internal roadmap vocabulary (`Slice 1.1`, `Phase B.5`) belongs in a shipped footer at all is a product question. Flagged, not answered.

**2. Every `placeholder=` is a judgement call, not only the one-word ones.**

#747 forced single-token placeholders to `placeholder` and left multi-word ones in `translate`. The triage showed that was the wrong cut — of six hits, **five were example values** that must not be translated (`Release sign-off`, `Release approver`, `id-1, id-2, …`, `email | uri | date-time | uuid`) and **one was a real instruction** (`Reason (optional)`). Nothing in the syntax separates an example value from an instruction, so multi-word placeholders now land in `review` — the bucket that means "a human decides", which is exactly true here.

## The translations

27 keys across six files, all of which reach **zero** actionable hits and join `SWEPT_COMPONENTS`.

The judgement calls worth naming, since that is what `review` exists to force:

- **`block` / `boost`** (kg-priorities action select) → German. GLOSSARY.md puts *descriptions of what is happening* in German and keeps *product nouns* English; these are actions an operator picks, so `blockieren` / `hervorheben`. The `value` attributes stay `block` / `boost` — the wire contract is untouched.
- **`Vault`, `Bucket`** → stay English (product noun, established loanword) but are now catalogue keys with explicit `brand` / `loanword` entries in `i18n-identical-allowlist.json`, rather than implicit exemptions.
- **`Status`** → identical in both locales, allowlisted as `loanword`.
- **`Resolution`** → `Behebung`. Not `Auflösung`, which in German software reads as screen resolution.

On that allowlist point: these five would have passed **silently** either way. `i18n-validate` exempts short capitalised nouns from the identical-value gate by design — a gap its own comment names as the follow-up. Listing them anyway puts the decision where a reader looks instead of relying on a rule that happens not to fire.

## Verification

- `npm run i18n:check` — OK, 3760 keys
- `tsc --noEmit` clean; `eslint` clean on every changed file
- `npm test` — **742/742 across 88 files**
- **Mutation check:** restoring one literal (`label={t('prefix')}` → `label="Prefix"`) turns exactly the new per-file ratchet red; reverting returns 15/15.

## What is left, and why

19 `translate` hits remain. Most are **sentences split across JSX elements** — the scanner sees `rendered markdown (`, `Stream live ·`, `· last activity`, `s tool-time`, `User · Tenant` — where the fix is an ICU message with placeholders, not a 1:1 key. Restructuring those changes markup rather than just extracting strings, so it belongs in its own pass rather than being smuggled into a translation PR.

The rest are taglines (`an Agentic OS`) and format names (`OpenAPI 3 (JSON)`) that need a brand decision before a key.

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/751"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789745253&installation_model_id=428086&pr_number=751&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F751&signature=f68163b1b24f565669c27529eddc39b54bb3d3ddbb851476cbb8ce1c3c756555"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/web-ui/app/_components/store/AdminUiPanel.tsx
+++ b/web-ui/app/_components/store/AdminUiPanel.tsx
@@ -8,6 +8,7 @@ import {
   useState,
 } from 'react';
 import { Globe } from 'lucide-react';
+import { useTranslations } from 'next-intl';
 
 import { cn } from '../../_lib/cn';
 
@@ -80,6 +81,7 @@ export function AdminUiArticleSwap({
   pluginName: string;
   children: React.ReactNode;
 }): React.ReactElement {
+  const t = useTranslations('store.adminUi');
   const { open, panelId } = useAdminUi();
   const ref = useRef<HTMLElement | null>(null);
 
@@ -108,7 +110,7 @@ export function AdminUiArticleSwap({
       </header>
       <iframe
         src={iframeSrc}
-        title={`${pluginName} Admin UI`}
+        title={t('iframeTitle', { plugin: pluginName })}
         className="h-[1000px] w-full rounded border border-[color:var(--rule)] bg-[color:var(--bg)]"
         loading="lazy"
       />
--- a/web-ui/app/_lib/i18n-structural.test.ts
+++ b/web-ui/app/_lib/i18n-structural.test.ts
@@ -130,7 +130,15 @@ describe('#679 / I3 — swept components carry no user-facing literals', () => {
   // So the ratchet now asks the scanner the same question the CLI asks:
   // does this file still contain a user-facing literal? Files join `SWEPT`
   // once they answer no, and can never silently regress afterwards.
-  const SWEPT_COMPONENTS = ['graph/_components/ListView.tsx'];
+  const SWEPT_COMPONENTS = [
+    'graph/_components/ListView.tsx',
+    'graph/_components/GraphCanvas.tsx',
+    'admin/duplicates/[id]/page.tsx',
+    'admin/duplicates/excerpt/[id]/page.tsx',
+    'admin/inconsistencies/[id]/page.tsx',
+    'admin/kg-priorities/page.tsx',
+    'system/_components/VaultStatusCard.tsx',
+  ];
 
   for (const file of SWEPT_COMPONENTS) {
     it(`${file} has no untranslated user-facing literal`, () => {
--- a/web-ui/app/admin/duplicates/[id]/page.tsx
+++ b/web-ui/app/admin/duplicates/[id]/page.tsx
@@ -124,11 +124,11 @@ export default function DuplicateDetailPage(): React.ReactElement {
               })}
             </p>
             <dl className="mt-3 grid grid-cols-[max-content_1fr] gap-x-3 gap-y-1 text-[11px]">
-              <dt className="text-[color:var(--fg-muted)]">Status</dt>
+              <dt className="text-[color:var(--fg-muted)]">{t('status')}</dt>
               <dd className="font-mono">{detail.props.status}</dd>
               {detail.props.resolution && (
                 <>
-                  <dt className="text-[color:var(--fg-muted)]">Resolution</dt>
+                  <dt className="text-[color:var(--fg-muted)]">{t('resolution')}</dt>
                   <dd className="font-mono">{detail.props.resolution}</dd>
                 </>
               )}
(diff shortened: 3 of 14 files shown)


Weegy:
Pushed a second commit that widens the scan — and it found three shipped **German** strings that the first version structurally could not see.

### Three blind spots, each a construct the parser never visited

**1. Template literals with substitutions.** `stringValueOf` handled plain string literals only, so this was invisible:

```tsx
title={`Triage-Klassifizierer: ${routing.classifierModel} → ${routing.bucket} → ${routing.model}`}
```

A hardcoded German tooltip in `chat/page.tsx`. Worth noting it also slips past the repo's own self-check in `web-ui/CLAUDE.md` — `git diff | grep '[äöüÄÖÜß]'` finds nothing, because "Triage-Klassifizierer" has no umlaut.

**2. Strings in plain objects rendered later.**

```tsx
const verdict = { simple: { label: 'einfach' }, complex: { label: 'komplex' }, … };
…
<span>{v.label}</span>
```

Neither JSX text nor a JSX prop. Two more German labels, same file. Restricted to the same closed prop list, so the only false positive it can produce is a `label` on an API payload — 24 hits repo-wide, 10 of them actionable, no explosion.

**3. My own over-exemption in the first commit.** It skipped **all** of `chat/page.tsx` because `web-ui/CLAUDE.md` exempts `MOCK_KG_WALK`. But the docs exempt a *symbol*, not a file — the fixture is a const at line 79, and the German tooltip sat at 1251. Exemptions are now scoped to the declaration. **An exemption must be no wider than the thing it excuses**, and that is the same over-broad-blessing failure this whole issue is about.

### Two corrections that fell out

- Rendering a template as `{}` per substitution **collided with the placeholder rule** (`\{[^}]*\}`), which would have marked every interpolated string exempt and silently re-hidden the class the change exists to reveal. Classification now runs on the literal parts; only the display form carries `{}`.
- Text with umlauts or ß is `translate` outright — German in a component is the severest class in the repo rules, so it outranks every exemption below. Precision over recall on purpose: it cannot false-positive, and it cannot catch umlaut-free German, which still lands in `review`. That is the honest limit; no cheap test separates the two languages, and guessing would put real English labels in the wrong bucket.

### The ratchet earned its keep

Widening the scan immediately turned `ListView.tsx` **red** — a `title={`User: ${user}`}` that was invisible when the file was declared swept in #747. That is the point of making the guard content-based rather than a list of strings: widening the question re-asks it of every swept file, instead of leaving old files blessed under the old, narrower question.

### Verification

`i18n:check` OK (3766 keys) · `tsc --noEmit` clean · eslint clean · **742/742 vitest**. Mutation check: restoring `label: 'einfach'` makes the object-label detector report it again; reverting returns 0.

PR #753, opened 2026-08-20
feat(ci): deterministic Odoo fixture + full verified/contradicted cov…

### LINKED ISSUE #639: feat(ci): golden-set eval v2 — deterministic-source + full-turn coverage (follow-up to #129)
Follow-up to #129 (golden-set regression eval, v1). Filed so the two deliberately-deferred coverage gaps in v1 do not get rediscovered later.

## Context — what v1 ships

The v1 golden set (#129) runs a frozen corpus through a **real** `VerifierPipeline` (real `ClaimExtractor` + `EvidenceJudge` on the pinned `VERIFIER_MODEL`) with **fixture-backed** deterministic sources and **no** Odoo/graph reader (`DeterministicChecker({})`). It asserts the verdict *class*. Corpus + runner live under `middleware/test/golden/`.

## Gap 1 — `approved` via a judge-*verified* soft claim is structurally unreachable in v1

Discovered while hardening the corpus. The stochastic stages run only when `shouldTriggerVerifier(answer)` fires, and it fires **only on a hard signal** (currency / date / accounting-ref / percent / duration). So:

- A soft claim with **no** signal → extraction skipped → `approved` before the judge runs (this is how v1's `approved` fixtures pass — trigger-skip only).
- Any answer that **does** trigger carries a hard claim, which `DeterministicChecker({})` resolves to `unverified` (no reader) → the turn lands in `approved_with_disclaimer`, never clean `approved`.

Net: **the judge's `verified → approved` path is never exercised in v1.** A regression where the judge stops confirming true claims (over-caution → everything disclaimered) is invisible. Closing this needs a deterministic source so a triggering answer's hard claim can resolve `verified`.

## Gap 2 — deterministic hard-claim verdicts are not covered

Same root cause: with no Odoo/graph reader, hard claims (amounts, refs, dates) can only ever be `unverified`. The deterministic-contradiction path (a hard claim the ERP data explicitly refutes → `blocked`) and the deterministic-confirmation path (→ contributes to `approved`) have zero corpus coverage.

## Gap 3 — full-turn generation eval

v1 evaluates the **verifier stage** on frozen `(userMessage, answer, trace)` tuples. It does not generate the answer. A model/prompt regression in the *orchestrator* (e.g. it stops calling a domain tool, per the calculator issue) is out of scope until a headless turn runner exists that can produce `input → live answer → verdict`.

## Proposed v2 scope

1. **Odoo/graph fixture reader** injected into `DeterministicChecker` (the `OdooReader`/`GraphReader` seams already exist). Enables:
   - a `verified` hard claim → genuine `approved` corpus entry (closes Gap 1),
   - a deterministic-contradiction `blocked` entry (closes Gap 2).
2. **Headless turn runner** so a fixture can be `(userMessage) → live orchestrator answer → verdict`, pinned to `ORCHESTRATOR_MODEL` / `SUB_AGENT_MODEL` (closes Gap 3). This is the larger lift and can be its own slice.
3. Extend `middleware/test/golden/` corpus + README with the new classes; keep the majority-of-3 flake policy and the key-gated `golden-eval.yml` trigger model.

## Acceptance criteria

- [ ] A corpus entry asserts `approved` through a judge/deterministic **verified** path (not trigger-skip).
- [ ] A corpus entry asserts `blocked` through a **deterministic** contradiction (ERP data refutes a hard claim), distinct from the judge-contradiction path.
- [ ] (Optional, may split out) A full-turn fixture drives `userMessage → live answer → verdict` against the orchestrator model.
- [ ] README documents the new fixture shapes and which model each new path pins.

Refs #129, #130, #131, #132.

sneumannb5:
## What

  Golden-set eval v2 (Closes #639). Inject FixtureOdooReader into DeterministicChecker so a corpus entry declares frozen Odoo records the real checker re-queries — closes Gap 1 (ERP-confirmed hard claim → verified → approved) and Gap 2 (ERP-refuted → blocked). expected.via asserts the path so an entry can't pass for the wrong reason. Gap 3 (full-turn eval) deferred.

## Test plan

  - [x] test/goldenRunner.test.ts (pure) — 33/33
  - [x] test/goldenModel.test.ts (real pipeline, stub LLM) — 9/9
  - [X] npm run lint && typecheck && test in middleware

## Risk / blast radius
 CI-only (test/golden/** + golden-eval.yml); no runtime/schema/API change, no new env-var.

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/753"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789798474&installation_model_id=428086&pr_number=753&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F753&signature=6710092ebea016a185c500dd3bac88fda8e461168777b9e67423dc3de3b10374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Weegy:
Reviewed and refreshed — `main` merged in at `619ab089` (the branch was based on `825dc714`, two merges behind, so its green was against a base that no longer exists).

### Verified on the new head

- `goldenRunner.test.ts` + `goldenModel.test.ts` — **42/42**
- full middleware suite — **6880 tests, 0 fail, 0 cancelled**

### The check I care about most passed

`expected.via` is a newly declared field, and the recurring defect in this repo is a field that nothing *reads* — it compiles, the tests pass, and the assertion is decorative. Not the case here: `viaSatisfied` is called at both decision sites, and the tests carry real negative cases (`viaPass === false` for the judge path, plus `throws on an unknown expected.via`).

Mutation-checked to be sure — forcing `viaSatisfied` to `return true` turns **2 tests red**, and reverting returns 33/33. The path assertion is load-bearing, not decorative.

### Cost, since this corpus is the one that just started actually running

I costed these evals recently, and got it wrong the first time by reading corpus size as call count. For anyone doing the same arithmetic here: the two new files are **2 entries**, not 70 — the rest of those line counts is (excellent) commentary.

The one thing worth naming explicitly, because "no new env-var" in the description is true but is not the whole cost story: folding `via` into the resample guard means an entry with the **right class but the wrong path** now resamples, where before it would have stopped after one call. `runEntry`'s own docstring already reasons this through correctly — green-on-first-sample costs one call, worst case stays bounded at `maxRuns × corpus size` — and the flake-tolerance argument for including `via` is sound. Flagging it only so the trade is visible in the PR, not just in the source.

No interaction with #748 (the `temperature` fix): the golden/verifier path sends no `temperature`, and `DEFAULT_MODEL` is `claude-haiku-4-5`, which accepts it regardless. Checked rather than assumed.

### One nit

The description says the blast radius is `test/golden/** + golden-eval.yml`, but `golden-eval.yml` is not in the diff — the workflow is untouched. Worth correcting so a reader does not go looking for a CI change that is not there.

No blockers from me.

Issue #937, opened 2026-08-28
OM-67: docs/CHANGELOG.md lists 144 released versions under [Unreleased]


Weegy:
Reported by Silvio Lange (TE Printline GmbH), omadia beta test round 3, 28.08.2026. Source: code analysis.

## Finding

In `docs/CHANGELOG.md` the `## [Unreleased]` section runs from line 19 to line 4419. The next version heading below it is `## [0.54.0] - 2026-07-06`. Everything shipped since then, 144 releases up to and including v0.140.3, sits under "unreleased".

Verified on `origin/main` at `5cc2f948`: `## [Unreleased]` at line 19, `## [0.54.0]` at line 4420.

## Cause

The file explains it itself in its header: the GitHub release notes are authoritative and the file is "a periodically-refreshed mirror". That refresh has not run since 06.07.2026 while new entries keep landing under `[Unreleased]`. The most recent one is #911, from the test day.

## Impact

Harmless internally, because we have the release notes. It is a trap for anyone reading the repository from outside. The reporter walked into it himself and classified OM-61 as "not yet released" although the fix had shipped that morning in v0.140.1, published at 09:40.

This is the only finding in the round that does not hit a user. It is filed because it affects the collaboration: if round 4 checks against the code again, the reporter needs a source he can trust.

## Proposed fix

Either make the mirror refresh run again, or add one sentence at the top of the file stating the version from which `[Unreleased]` no longer means anything.

Issue #935, opened 2026-08-28
OM-59: Electron dialogs, loading screen and menu headings stay English while the UI is German


Weegy:
Reported by Silvio Lange (TE Printline GmbH), omadia beta test round 3, 28.08.2026. Build v0.139.1, macOS arm64, UI language set to "DE - DEUTSCH".

## Symptom

With the UI language set to German:

- "omadia 0.140.1 is ready to install. omadia will close, back up your local data, and restart to apply the update." with *Later* / *Restart now*
- "omadia could not start its local services. Error: boot superseded." with *Quit* / *Re-run setup*
- Loading screen: "Waiting for the kernel to become healthy..."

## Cause: three language layers that do not know about each other

| Layer | Language source | State |
|---|---|---|
| Web UI | `next-intl`, own switcher (`DE - DEUTSCH`) | fully translated |
| Setup wizard | `wizard-i18n.js`, OS language (`navigator.language`) | translated |
| Electron dialogs + loading screen | none | English throughout |

The third row is the finding. The dialog strings are hardcoded in `desktop/src/updater.ts` and `desktop/src/main.ts`, `loading.html` is wired in English, and `loading.js` renders `p.message` raw, i.e. the English supervisor text.

The annoying part is that the German strings already exist. `wizard-i18n.js` carries `boot.starting-db`, `boot.starting-kernel`, `boot.waiting-kernel`, `boot.starting-ui`, `boot.ready` in German, and `wizard.js` uses them. Only the loading screen, which every already-configured user sees on every start, does not reach for them. So a first-time user gets German boot messages and a daily user gets English ones.

The app menu also carries hardcoded English headings in `desktop/src/menu.ts`: "File", "Edit", "View", "Window", "Help". Electron translates the entries underneath via their `role:` values, but not the headings.

## Impact

This is OM-50 from round 2 in a new and more sensitive place. The user is asked to decide in English whether his program shuts down and touches his data. "boot superseded" is pure developer language on top of that, and even translated the term would be empty for a business user (see OM-56).

The web UI itself is translated well by now. The desktop shell dialogs are not.

## Proposed fix

- Smallest effective step: let `loading.js` use the same `wt()` function as `wizard.js`. The dictionary entries are already there.
- Then route the four dialog strings in `updater.ts` and `main.ts` through the same dictionary.
- Localize the menu headings in `menu.ts`.

Issue #926, opened 2026-08-28
OM-55: updater downloads the same version three times, never installs it, and logs no error


Weegy:
Reported by Silvio Lange (TE Printline GmbH), omadia beta test round 3, 28.08.2026. Tested build: v0.139.1, desktop app macOS arm64.

## Symptom

Within nine minutes omadia offers version 0.140.1 three times. After every run the machine is still on 0.139.1. At the end of the session the app does not start at all any more.

Log (UTC):

```
10:07:31 [updater] update available: 0.140.1
10:09:48 [updater] downloaded 0.140.1
10:11:17 [updater] snapshotted DB -> .../snapshots/pgdata-pre-0.140.1
10:14:11 [updater] update available: 0.140.1
10:14:11 [updater] downloaded 0.140.1
10:15:13 [updater] snapshotted DB -> .../snapshots/pgdata-pre-0.140.1
10:15:25 [updater] update available: 0.140.1
10:15:25 [updater] downloaded 0.140.1
10:16:10 [updater] snapshotted DB -> .../snapshots/pgdata-pre-0.140.1
```

No error line anywhere. `desktop/src/updater.ts` would log `pre-install stop/snapshot failed` on a failing stop or snapshot, and `autoUpdater.on('error')` would catch an updater error. Neither fires. The snapshot succeeds every time and `quitAndInstall()` is reached.

`initUpdater()` calls `checkForUpdates()` once per app start, so three cycles mean three full app restarts and three installer failures.

## Cause

Kernel and embedded Postgres survive the quit and keep running out of the app bundle that ShipIt is supposed to replace:

```
5159  .../omadia.app/Contents/Resources/omadia-pg/bin/postgres -D .../pgdata
5168  .../omadia.app/Contents/MacOS/omadia .../middleware/dist/index.js
6374  .../omadia.app/Contents/Frameworks/Squirrel.framework/Resources/ShipIt
```

ShipIt was still waiting more than an hour later. `/Applications/omadia.app` reports 0.139.1 with an mtime from the previous day, so not a single write attempt landed. The bundle is owned by the logged-in user with `rwx`, so permissions are not the cause.

The root cause is that `Supervisor.stop()` can return while children are still alive. Tracked as OM-64.

[...]

Issue #925, opened 2026-08-28
OM-62: PATH augmentation misses ~/.local/node/bin, so the subscription CLI install cannot find npm


Weegy:
Reported by Silvio Lange (TE Printline GmbH) in the omadia beta test, round 3, 28.08.2026. Tested build: v0.139.1, desktop app macOS arm64 (`ai.omadia.desktop`).

**This is the top-priority finding of the round.** Nothing else in the product is testable through the desktop app until the subscription path works, because a normal user has no API key with credit.

## Symptom

Admin -> LLM access -> Subscriptions -> "Install now" fails with:

```
npm install failed - no output at all; npm was most likely not found.
```

npm and Node are installed on the machine:

```
$ which npm; node -v
/Users/silviolange/.local/node/bin/npm
v24.19.0

$ ls ~/.local
node
```

There is no `~/.local/bin` on that machine. Only `~/.local/node`.

## Cause

`desktop/src/pathEnv.ts` builds `candidateDirs` from a fixed list and then drops everything that does not exist (`existingDirs`). The list contains `~/.local/bin` (added in #906 / PR #907 for Claude Code's native installer) but not `~/.local/node/bin`, so this npm is never on the augmented PATH.

The error message maps cleanly to `middleware/src/platform/cliInstallService.ts:178-179`, the `cli_install.no_output` branch, which fires exactly when npm produced no output at all.

Verified against `origin/main` at `5cc2f948`: the fixed-path list and the `existingDirs` filter are unchanged.

## Second hole of the same kind

`resolveNvmBinDir()` reads `~/.nvm/alias/default` and accepts the value only if `isLiteralNodeVersion()` matches, i.e. a bare version number. The two most common nvm defaults, `lts/*` and `node`, both fail that test, so nvm users with either setting land in the same error. Not reproduced on a device, but it is the same class of bug and belongs in the same fix.

## Proposed fix

- Scan one level below `~/.local` for `*/bin` instead of probing fixed single paths. That covers `~/.local/node/bin`, `~/.local/bin` and equivalents in one rule.
- Resolve the nvm `default` alias instead of discarding it, so `lts/*` and `node` work.
- Include the PATH that was actually searched in the "details for support" panel, so a user can see where omadia looked.

## Acceptance

- A machine with npm only in `~/.local/node/bin` completes the subscription CLI install.
- A machine with `~/.nvm/alias/default` set to `lts/*` resolves the nvm bin dir.
[...]

PR #943, opened 2026-08-28
fix: promote a release to latest only once its artifacts exist

### LINKED ISSUE #928: OM-69: macOS update feed 404s for 30-60 minutes after every release
Reported by Silvio Lange (TE Printline GmbH), omadia beta test round 3, 28.08.2026. Found during a re-test on the running system at 14:36.

## Symptom

Tray menu -> "Check for Updates..." on a machine running 0.139.1:

> **omadia could not check for updates.**
> Error: Cannot find `latest-mac.yml` in the latest release artifacts
> (`https://github.com/byte5ai/omadia/releases/download/v0.141.0/latest-mac.yml`): **HttpError: 404**

The machine is not offered an outdated version. It is offered nothing at all.

## Cause

- v0.141.0 was tagged at 14:08:52 (`5cc2f948`, #923), 28 minutes before the check.
- `.github/workflows/auto-release.yml:119` creates the release with `gh release create "$new" ... --latest`.
- The job that builds and attaches the macOS artifacts runs after that: `desktop-apps` with `needs: release`.
- `desktop-apps.yml` documents its own runtimes: linux 5 min, windows 9 min, macOS arm64 21 min, macOS Intel over 30 min. `latest-mac.yml` is only produced once both architectures are done and `merge-mac-update-feed.mjs` has merged them, plus notarization time, which Apple controls.
- `electron-builder.yml` publishes with `provider: github`, `releaseType: release`, so electron-updater always queries whatever release is flagged "latest", which is the one that has no artifacts yet.

Between `gh release create --latest` and the end of the desktop build, every macOS install queries a file that does not exist. Realistically a 30 to 60 minute window.

## Impact

This is not an edge case, it is every release. 144 releases since 06.07.2026, five on the test day alone (v0.140.0 through v0.141.0). In each of those windows the built-in update is dead for all macOS users. If the desktop build fails outright, the channel stays dead until the next successful release.

The worse half is silence. The reporter only saw the error because he checked manually. On the silent check at app start the same error goes to the log only, because the dialog is gated on `takeManualCheckPending()`. A user gets neither an update nor a hint that anything is wrong.

## Proposed fix

Gate the "latest" flag on the artifacts actually being present:

- Create the release with `--draft`, or without `--latest`.
- After the `desktop-apps` job succeeds, promote it with `gh release edit "$new" --latest`.

Then the updater only ever sees releases it can actually install.

Additionally: a 404 on the feed file is a different state from "no new version". A repeated silent failure should surface once, otherwise a dead update channel is indistinguishable from "I am up to date" (same mechanism as OM-52 from round 2).

Weegy:
Closes #928

Beta-test round 3 finding OM-69: after every release the macOS update channel is dead for 30-60 minutes, and it never says so.

## Part 1 — two-phase publish

`auto-release.yml` created the release with `gh release create ... --latest`, and the `desktop-apps` job that attaches the installers runs *after* it (`needs: release`). `latest-mac.yml` only exists once **both** mac architectures finish and `merge-mac-update-feed.mjs` has merged them: arm64 21 min + Intel >30 min by the workflow's own measurements, plus notarization on Apple's clock. In that window every macOS install queries a file that does not exist and gets a 404. 144 releases since 2026-07-06, five on the test day alone.

Now:

1. `release` creates the release with `--draft` (no `--latest`).
2. New `promote-release` job runs `gh release edit "$TAG" --draft=false --latest` once the merged macOS feed **and the installers it references** are attached.

### Why a draft rather than just clearing `make_latest`

**Correcting an earlier version of this PR description, which had this wrong.** I first claimed electron-updater resolves "latest" from the first entry of `releases.atom`, making `--latest=false` useless. That is not what the code does. With `allowPrerelease` false, `GitHubProvider.getLatestVersion()` takes the `else` branch and resolves the version via `getLatestTagName()` → `GET <repo>/releases/latest` with `Accept: application/json`, reading `tag_name` — and that endpoint **does** honour `make_latest`. The atom feed is fetched too, but only to locate the matching entry for the release notes. `allowPrerelease` is false here because `AppUpdater` derives it from the running version via `hasPrereleaseComponents()` and `set-desktop-version.mjs` writes plain semver (`VERSION_RE` enforces `vX.Y.Z`); nothing in `desktop/src/updater.ts` or `electron-builder.yml` overrides it.

So `--latest=false` would have fixed OM-69 on its own. I got this wrong by grepping for the atom fetch and mistaking the notes-lookup path for the version-selection path, without reading the `if/else`.

The draft approach stands, and is strictly stronger in two ways that matter:

1. A draft is absent from **both** `/releases/latest` and `releases.atom`, so the protection does not depend on which the provider consults.
2. That choice is conditional on `allowPrerelease`, which flips true under any prerelease version scheme — activating the first-atom-entry path where `make_latest` is ignored entirely, silently removing the protection. A draft is invisible either way.

A draft also keeps a half-built release off the Releases page while it is unusable.

### The gate proves downloadability, not just a feed

A valid merged feed does **not** prove the installers it points at are on the release. The feed is assembled from **run artifacts**, uploaded by an `if: always()` step that runs *before* the release upload. So a mac leg whose "Upload installers to the release" step failed transiently (API 5xx) still yields a complete, mergeable feed — the merge succeeds, the feed uploads, the release is flagged latest, and `latest-mac.yml` references assets that were never attached. OM-69's exact symptom, under a green run.

The gate now asserts every `path`/`url` in the merged feed against the release's actual asset list before setting `published=true`, reusing the strict `parseFeed()` from `merge-mac-update-feed.mjs` and decoding the percent-escaping electron-builder applies. Tested both directions locally (all-attached passes with correct dedupe of `path` against `files[].url`; one missing installer fails with the name).

### Gating on the output, not the job result

The promotion hangs off `desktop-apps`' `mac_feed_published` output rather than `needs.desktop-apps.result`. `build` runs `fail-fast: false`, so a broken **Windows** leg fails the called workflow while the macOS feed published fine — gating on the result would withhold the promotion and leave macOS auto-update dead until the next fully green release, which is OM-69 rather than a fix for it.

### Failure and cancellation semantics (deliberate)

| Situation | Result |
|---|---|
| feed + referenced installers attached | release published + flagged latest |
| `desktop-apps` fails, feed and installers attached anyway (e.g. Windows leg only) | release **still promoted** — macOS can update |
| feed published but an installer missing | release **stays a draft**, job fails **red** |
| feed not published (mac leg failed, cancelled, skipped) | release **stays a draft**, job fails **red** |
| no release cut (no feat/fix) | all jobs skipped, unchanged |

When it stays a draft the previous release keeps the latest flag, so installs keep updating to the last version they can actually download. Recovery: re-run `desktop-apps`, or `gh release edit <tag> --draft=false --latest`.

Two side effects of leaving a stuck draft are now documented on the job: `publish-images` needs only `release`, so the container `:latest` tag ships a version desktop clients cannot get; and the changelog range comes from `git tag -l` with the tag pushed before the release exists, so the next release's notes start *after* the stuck tag and those commits appear in no published notes until the draft is published.

### Draft-aware, paginated release lookup

`GET /releases/tags/{tag}` does not return drafts, so both upload steps resolve the release through the list endpoint instead.

**A second correction:** I originally justified this by claiming `gh release view` cannot resolve drafts. It can — gh's `FetchRelease` finds a published release by tag *or* a draft by its pending tag name, with a GraphQL fallback on 404. The trap I described did not exist. I kept the replacement anyway because `require_release` is genuinely valuable, and that same fallback is why `gh release upload "$TAG"` works against a draft at all — the single assumption this whole flow depends on, now positively confirmed rather than assumed, and stated in a comment.

The lookups use `--paginate`, which is not optional: the repo has **237** releases and one `per_page=100` page reaches back only to **v0.78.0**. Without it, a rebuild dispatched for an older tag found nothing and — on the dispatch path, where `require_release` is false — took the soft-skip branch and reported a green run that attached nothing. That was a regression this PR introduced by replacing `gh release view`, which resolved any tag regardless of age.

`gh`'s built-in gojq is used rather than piping to `jq`, because the installer step also runs on `windows-latest` where `gh` is guaranteed and `jq` is not.

### No re-trigger loop

`desktop-apps.yml` also triggers on `release: [published]`. Promoting with `GITHUB_TOKEN` does not fire that event (Actions loop protection — the same reason `publish-images` and `desktop-apps` are called in-run here, as this workflow's header already documents). A comment warns against "fixing" this with a PAT.

## Part 2 — a dead channel now says so

The reporter only saw the 404 because he checked manually. On the silent startup check the same error goes to the log only, since the dialog is gated on `takeManualCheckPending()`. A failed check and "you are up to date" were indistinguishable.

`desktop/src/updaterCheckHealth.ts` (new) tracks consecutive failures across app starts in Electron's `userData` dir — updater bookkeeping, not user data, so it survives the user moving their data directory. One warning per failure streak; any check that gets through resets it. Threshold is 3 app starts (`initUpdater()` runs one check per start), so a single blip cannot produce a dialog.

The reset fires on `update-downloaded`, **not** `update-available`. `autoDownload` is on, so a download follows announcement immediately and its failures arrive on the same `error` event; resetting on announcement read "announce → reset to 0, download fails → 1" every start, capping the counter at 1 so the warning could never fire for a broken download — the residual case the asset assertion above cannot cover. `update-not-available` keeps its reset for the healthy path.

The decision is a pure function (`nextFailureState`) kept apart from the file IO, so the notify-once-per-streak rule is unit-testable without an Electron runtime.

## Verification

```
cd desktop && npx tsc -p tsconfig.json --noEmit   -> clean
cd desktop && npm test                            -> 33/33 pass
actionlint auto-release.yml desktop-apps.yml      -> 3x SC2086, all pre-existing
                                                     (identical to the same lint on HEAD)
```
The feed-vs-assets assertion was exercised locally in both directions, including percent-decoding and the relative `import()` resolving from the repo root as it will in the job. Confirmed live: 237 releases with page 1 ending at v0.78.0; `/repos/byte5ai/omadia/releases/latest` returns the `make_latest`-flagged release.

## Notes

- **`docs/CHANGELOG.md` deliberately untouched** — it is the universal merge-conflict point across the round-3 cluster PRs; one consolidated entry lands with #937.
- **Watch the first release after this merges.** The draft path is new; a misbehaving promotion leaves the release a draft and the run red, which is the safe direction but needs someone to see it.
- Known, accepted: one transient failed **manual** check sets `noticeShown` and suppresses the passive warning for the rest of that streak, so a user who clicks "check for updates" during a blip loses the warning. It resets on the next successful check, so it is not a wedge.
- New dialog strings are English, consistent with the rest of `updater.ts`; #935 localizes the desktop-shell dialogs as a set.
- Scope kept off `supervisor.ts`, `main.ts`, `paths.ts`, `ipc.ts`, `renderer/**`, `menu.ts`, `middleware/**`, `web-ui/**`, which other round-3 clusters own. Re-checked against #944 (now 21 files, `updater.ts` +161/-13): still no semantic overlap — it adds and removes nothing in the `error` handler, the `takeManualCheckPending()` gate, or the `update-*` paths that constitute my change. Expect a mechanical conflict in the import block, and now also at the `update-downloaded` opening line, since both add a statement after its `log.info`.
- `omadia-reviewer` was not run against this diff (subagents cannot spawn agents).

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -30,9 +30,11 @@ name: auto-release
 #   2. That choice is conditional on `allowPrerelease`, which AppUpdater derives
 #      from the running version (`hasPrereleaseComponents`). It is false only
 #      because set-desktop-version.mjs writes plain semver. Adopt any prerelease
-#      version scheme and it flips true, activating the first-atom-entry path
-#      where `make_latest` is ignored entirely — and this protection would
-#      silently stop working. A draft is invisible either way.
+#      version scheme and it flips true, moving version selection onto the atom
+#      feed, where `make_latest` is ignored entirely — and this protection would
+#      silently stop working. (Which atom branch depends on the channel: with no
+#      channel it takes the first entry, with one it scans entries for a match.
+#      Either way the flag is not consulted.) A draft is invisible either way.
 # A draft also keeps a half-built release off the Releases page while it is
 # unusable, rather than showing one nobody can install.
 #
@@ -200,7 +202,15 @@ jobs:
   # Failure semantics, deliberately chosen: if the feed never published, the
   # release STAYS A DRAFT and this job fails red. The previous release keeps the
   # latest flag, so existing installs keep updating to the last version that can
-  # actually be downloaded; nothing half-published ever becomes latest. The cost
+  # actually be downloaded; no release lacking its macOS feed becomes latest.
+  #
+  # Note the limit of that promise: this gate is macOS-only, by the same design
+  # decision as gating on the feed output rather than the job result. A red
+  # WINDOWS leg still promotes, so `latest.yml` can be missing while the release
+  # is flagged latest, and Windows clients then hit OM-69's 404 on their own feed.
+  # That is a deliberate trade — withholding promotion on any red leg would leave
+  # macOS auto-update dead until the next fully green release — not a guarantee.
+  # Extending the assertion to `latest.yml` is the obvious follow-up. The cost
   # is that the new release notes stay invisible until someone acts, which is why
   # this job fails loudly instead of exiting 0 — a red run is the signal. Recover
   # by re-running the desktop-apps job, or publish by hand once the artifacts are
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -7,16 +7,34 @@ name: auto-release
 #
 # TWO-PHASE PUBLISH (#928). The release is created as a DRAFT and only promoted
 # to published+latest once `desktop-apps` has attached the merged
-# `latest-mac.yml`. It has to be a draft, not merely `--latest=false`:
-# electron-updater's GitHubProvider resolves "latest" by fetching
-# `<repo>/releases.atom` and taking the FIRST entry, and that feed is ordered by
-# publication with no notion of GitHub's `make_latest` flag — so an unflagged
-# but published release is still the one every client queries. Drafts are absent
-# from the feed entirely, which is the only thing that actually holds clients on
-# the previous release. Before this, every release opened a 30-60 minute window
-# (macOS arm64 21 min + Intel >30 min + Apple-controlled notarization) in which
-# every macOS install asked for a `latest-mac.yml` that did not exist yet and got
-# a 404 — 144 releases since 2026-07-06, five on one day.
+# `latest-mac.yml` AND the installers that feed references.
+#
+# The problem: `desktop-apps` runs `needs: release`, and `latest-mac.yml` only
+# exists once both mac architectures finish and merge-mac-update-feed.mjs has
+# combined them — macOS arm64 21 min + Intel >30 min by that workflow's own
+# measurements, plus notarization on Apple's clock. Flagging the release latest
+# up front therefore opened a 30-60 minute window per release in which every
+# macOS install asked for a file that did not exist yet and got a 404. 144
+# releases since 2026-07-06, five on one day.
+#
+# Why a DRAFT rather than just clearing `make_latest`: clearing the flag alone
+# would in fact work today. electron-updater (6.8.9, GitHubProvider) resolves the
+# version through `getLatestTagName()` -> `GET <repo>/releases/latest` with
+# `Accept: application/json`, reading `tag_name` — and that endpoint honours
+# `make_latest`. It also fetches `releases.atom`, but in our configuration only to
+# locate the matching entry for the release notes.
+#
+# A draft is preferred because it is strictly stronger, in two ways that matter:
+#   1. It is absent from BOTH `/releases/latest` and `releases.atom`, so it does
+#      not depend on which of the two the provider happens to consult.
+#   2. That choice is conditional on `allowPrerelease`, which AppUpdater derives
+#      from the running version (`hasPrereleaseComponents`). It is false only
+#      because set-desktop-version.mjs writes plain semver. Adopt any prerelease
+#      version scheme and it flips true, activating the first-atom-entry path
+#      where `make_latest` is ignored entirely — and this protection would
+#      silently stop working. A draft is invisible either way.
+# A draft also keeps a half-built release off the Releases page while it is
+# unusable, rather than showing one nobody can install.
 #
 # HARD RULE: every tagged release gets a clean, categorized changelog with no
 # manual step required — enforced by construction, since it's part of the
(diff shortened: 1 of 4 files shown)


Weegy:
Rebased-by-merge onto `main` @ `510a681f` (picks up #941 / issue #925). Clean merge, no conflicts.

Re-verified on the new base:

```
cd desktop && npx tsc -p tsconfig.json --noEmit   -> clean
cd desktop && npm test                            -> 33/33 pass  (22 before; #941's pathEnv tests now run)
actionlint auto-release.yml desktop-apps.yml      -> 3x SC2086, all pre-existing
```

**Relationship to #944** (issues #926/#934, also open, `desktop/src/updater.ts` +140/-12): checked hunk by hunk, the two do not overlap semantically. #944 adds and removes nothing in the `autoUpdater.on('error')` handler, the `takeManualCheckPending()` gate, or the `update-available` / `update-not-available` paths, which is the entirety of my `updater.ts` change. Its nearest hunk sits in the import block and the `autoUpdater` setup lines just above my edit, so whichever merges second gets a mechanical two-independent-added-lines conflict at most.

The two persistence modules are also distinct concerns despite similar names: #944's `updateAttempts.ts` records *install* attempts (did the expected version come back after `quitAndInstall`), mine records *check* failures (did the feed answer at all). Worth a follow-up to consider whether they should share one state file, but not worth coupling the PRs now.

PR #1121, opened 2026-09-18
docs(channel-api): align public API README with actual route behavior…

### LINKED ISSUE #1110: Public API README is out of date: done example omits four fields, the AI-disclosure paragraph appears only once per conversation, 429 has no Retry-After, and the same key gets a second rate-limit budget on the MCP route
### What happened?

Five points where `middleware/packages/harness-channel-api/README.md` does not
match what the route actually does in v0.159.2. Each one is small; together they
mean an integrator who builds strictly against the README ends up with a client
that misbehaves in ways the README says are impossible.

The code is right in all five cases. The README is what needs changing.

### 1. The `done` example omits four fields the route actually sends

`README.md:251` shows:

```
{"type":"done","answer":"Your current MRR is...","toolCalls":0,"iterations":1}
```

The real `done` event also carries `provenance` (the AI Act Article 50 envelope
marker, stamped by this plugin), `aiDisclosure` (the structured disclosure
object), `runTrace` and `palaiaExcerpt`. A client written against the shown
shape will not know that `provenance` and `aiDisclosure` exist, which matters
because those two are exactly the fields a compliance-relevant integration
needs. Please extend the example, or add a field table for `done` next to the
event-type table.

### 2. The AI-disclosure paragraph appears only in the first turn of a conversation, and this is documented nowhere

The disclosure line ("Diese Antwort wurde von einem KI-System erzeugt.") is
folded into the answer text only on the first turn of a conversation scope. On
every later turn with the same `conversationId` the text no longer contains it,
while the structured `aiDisclosure` field keeps riding every turn. That split is
deliberate and correct, see the fold-once rationale in
`middleware/packages/harness-channel-sdk/src/aiDisclosure.ts`, but the README
never mentions it.

Two consequences an integrator will hit:

- The same request sent twice produces answer text that differs in a whole
  paragraph, which looks like nondeterminism.
- Without a `conversationId`, every call is its own scope, so every call folds
  the paragraph in again.
- The seen-store is in memory and per process, so a restart makes the next turn
  in an existing conversation fold the paragraph in once more.

Please state in the README that `aiDisclosure` is the stable carrier and must be
read from the structured field, and that the presence of the paragraph inside
`done.answer` is not something to depend on.

### 3. GET on the chat route is not mentioned

`GET /api/public/v1/chat` returns `404`, which is correct, the router only
registers POST. The README never says the route is POST-only, so one sentence in
the request section would save someone a debugging round.

### 4. The retry advice promises a backoff signal that is not sent

`README.md:140-142` and `:215` tell the caller to back off and retry after the
window resets. The `429` response
(`middleware/packages/harness-api-key-auth/src/requireApiKey.ts:113-119`) carries
no `Retry-After` header, so a client has no machine-readable way to know when
the window resets and has to hardcode 60 seconds from prose. Either send
`Retry-After`, or say plainly in the README that no such header exists and that
the window is a fixed 60 seconds.

### 5. The same API key gets a separate rate-limit budget on the MCP route, and the README does not say so

The same key is also valid for `POST /api/v1/mcp`. Both routes call
`createRateLimiter()` independently
(`middleware/packages/harness-channel-api/src/plugin.ts:58` and
`middleware/src/mcp/wirePublicMcp.ts:340`), so one key's effective budget is
`rateLimitPerMinute` per route, not per key. The README's rate-limiting section
describes the budget as if it were per key. An operator sizing a key for a
customer will get twice the traffic they planned for.

### Expected

The README describes the route as it behaves: the full `done` shape, the
fold-once disclosure rule with `aiDisclosure` named as the stable carrier,
POST-only, no `Retry-After`, and a per-route rather than per-key budget.

### Steps to reproduce

Prerequisites: a running omadia instance with the `@omadia/channel-api` plugin
active, and an API key with the `chat:write` scope and a low
`rateLimitPerMinute` (5 is convenient for point 4).

```bash
export API="http://localhost:8080"
export KEY="omk_<your-key>"
```

### 1. `done` shape versus the README example

```bash
curl -sS -N -X POST "$API/api/public/v1/chat" \
  -H "authorization: Bearer $KEY" \
  -H 'content-type: application/json' \
  -d '{"message":"Answer with pong only.","conversationId":"doc-check-1"}' \
  | grep '"type":"done"' | python3 -m json.tool
```

Compare the keys against `README.md:251`. `provenance`, `aiDisclosure`,
`runTrace` and `palaiaExcerpt` are present on the wire and absent from the
example.

### 2. Disclosure paragraph only in the first turn

Send the exact same call twice with the same `conversationId`:

```bash
for i in 1 2; do
  curl -sS -N -X POST "$API/api/public/v1/chat" \
    -H "authorization: Bearer $KEY" \
    -H 'content-type: application/json' \
    -d '{"message":"Answer with pong only.","conversationId":"doc-check-2"}' \
    | grep '"type":"done"'
done
```

The first `done.answer` contains the disclosure paragraph, the second does not.
`aiDisclosure` is present in both. Repeat without `conversationId` and every
call contains the paragraph.

### 3. GET on the chat route

```bash
curl -sS -i "$API/api/public/v1/chat" -H "authorization: Bearer $KEY"
```

Returns `404`, which is correct and undocumented.

### 4. No Retry-After on 429

Send one more call than the key's per-minute budget allows within one minute and
inspect the headers of the rejected response:

```bash
curl -sS -i -X POST "$API/api/public/v1/chat" \
  -H "authorization: Bearer $KEY" \
  -H 'content-type: application/json' \
  -d '{"message":"ping"}' | head -20
```

The `429` response has no `Retry-After` header.

### 5. Separate budget on the MCP route

With the same key, exhaust the budget on `/api/public/v1/chat` until a `429`
arrives, then immediately call `POST /api/v1/mcp` with the same key. The MCP
call is not rate limited, because it counts against its own limiter instance.

### omadia version

v0.159.2

### How are you running it?

Docker Compose (prebuilt GHCR images)

### Relevant logs

```shell
`429` response, note the absent `Retry-After`:


{"error":"rate_limited","message":"this key is limited to 5 requests/minute"}


Relevant code:

- `middleware/packages/harness-channel-api/README.md:140-142`, `:215`, `:251`
- `middleware/packages/harness-channel-sdk/src/aiDisclosure.ts` — fold-once-per-scope rule and the in-memory seen store
- `middleware/packages/harness-api-key-auth/src/requireApiKey.ts:113-119` — `429` without `Retry-After`
- `middleware/packages/harness-channel-api/src/plugin.ts:58` and `middleware/src/mcp/wirePublicMcp.ts:340` — two independent `createRateLimiter()` instances
```

sneumannb5:
## What
  Fix five spots where `middleware/packages/harness-channel-api/README.md` no longer matched the route's actual behavior (v0.159.2). Docs-only — the code is correct in all five cases (Closes #1110).

  ## Why
  An integrator building strictly against the README ended up with a client that misbehaves exactly where the docs said it couldn't — including the two compliance-relevant fields (`provenance`, `aiDisclosure`). The README is now accurate to the wire.

  - `done` field table: `provenance`, `aiDisclosure`, `runTrace`, `palaiaExcerpt` (was: 3-field example only)
  - fold-once-per-scope disclosure rule documented; `aiDisclosure` named as the stable carrier to read from
  - route is POST-only (`GET` → 404)
  - no `Retry-After` on 429; window is a fixed 60s
  - rate-limit budget counted per limiter, not pooled across routes (chat + MCP each get their own)

  ## Test plan
  - [x] no code touched — docs-only, no lint/typecheck/test surface
  - [x] every documented field/behavior verified against source (`aiDisclosure.ts`, `provenance.ts`, `chatRouter.ts`, `requireApiKey.ts`, `plugin.ts`, `wirePublicMcp.ts`)
  - [x] two-axis review (Standards + Spec): 0 hard violations, 5/5 spec points covered

  ## Risk / blast radius
  Documentation only — no schema, no code, no CI, no env-var. Public API *docs* now describe existing behavior more accurately; no wire change. Two behavior improvements deferred to follow-up issues (429 `Retry-After` header; unify chat+MCP to a per-key budget).

---
<a href="https://app.blacksmith.sh/byte5ai/codesmith/omadia/pr/1121"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792304372&installation_model_id=428086&pr_number=1121&repository=byte5ai%2Fomadia&return_to=https%3A%2F%2Fgithub.com%2Fbyte5ai%2Fomadia%2Fpull%2F1121&signature=6ebf2a5cb71a2a3b759164bb9606f0079d08e8cd5260adbf3a0f5f0794383996"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Weegy:
Superseded by #1164, which carried your commit plus a review pass and is now merged (6b33d675); #1110 is closed by it. Main had moved since this PR (#1122, #1126, #1137, #1138), so the README additionally reconciles the fold-once disclosure rule, the MCP write sub-cap, and the new `answerSource`/`receiptId` fields, and the POST-only / no-Retry-After contracts are now pinned by tests. Pushing to the fork branch was not available from the review session, hence the separate PR. Thanks!

